PR #54, opened 2025-04-17
[Analyze 1886] Fix PA condition occurence id filter

brandantck:
- resolves: https://github.com/data2evidence/internal/issues/1886

1. Fix condition occurenence id type
2. Add translation for `LIKE_REGEXPR` in duckdb->hana translation as `LIKE_REGEXPR` is not a valid function in duckdb

### Merge Checklist


p-hoffmann (approved):

suwarnoong (approved):

functions/_shared/alp-base-utils/src/helpers/hanaTranslation.ts
```diff
@@ -211,6 +211,9 @@ export const translateHanaToDuckdb = (
   );
 
   temp = hanaCommonTranslation(temp, schemaName, vocabSchemaName);
+
+  // Cast left comparator to varchar which is required by duckdb
+  temp = temp.replace(/LIKE_REGEXPR/gi, "::VARCHAR ILIKE");
```
satish-a0, line 216: LIKE_REGEXPR has never been used in the Duckdb cdw config. Instead `JARO_SIMILARITY` is being used.

brandantck: The `LIKE_REGEXPR` should be coming from source code when the `type` in cdm config is set to `text`.
I might be wrong but i think its from [here](https://github.com/OHDSI/d2e/blob/develop/functions/query-gen-svc/src/svc/DomainValuesSvc.ts#L148).
For postgres translations there was an [existing translation](https://github.com/OHDSI/d2e/blob/develop/functions/_shared/alp-base-utils/src/helpers/hanaTranslation.ts#L183) for `LIKE_REGEXPR` so i suppose it was to handle cases like this, so i added a translation in duckdb as well

SantanM (approved):

functions/_shared/alp-base-utils/src/helpers/hanaTranslation.ts
satish-a0: If that was the case drop down values attribute of domain value service wouldnt be working already in test system now. Because its not used and is taken from cdw config.

brandantck: I just tried on the demo system, using the `condition occurence id` filter will throw an error
image

brandantck: Test system as well throws error if we use `condition occurence id`
image
image

brandantck: Sorry i forgot to turn off the auto-merge so the PR was merged before this comment was resolved

satish-a0: If you go and update type to num in cdm config, the error would go away. LIKE_REGEXPR and the above error are two separate things in PA, its fine.

brandantck: Hmm yes, so the fix in the PR was to fix the type to `num`, and also add the `LIKE_REGEXPR` translation as a failsafe. Was thinking since the postgres translation has it, its also better for it to be added for the duckdb translation

brandantck: Do you happen to know where that `LIKE_REGEXPR` is coming from? When type is set to `text`

PR #873, opened 2025-09-14
mv function env to pkg json

p-hoffmann:
### Merge Checklist


functions/package.json
```diff
@@ -287,6 +287,283 @@
           "rmsrc": true
         }
       ],
+      "env": {
```
satish-a0, line 290: Could you elaborate the reason for moving the envs here. Since it involves an additional layer of processing by trex and building of function specific packages if there's a change in the envs now.

p-hoffmann: The docker compose file should be easimy to understand by people who are running d2e and we should start documenting the env variables in it.

A change to the package.json is only required if the env variable has a hard coded value. For env variables which can change we already use variables, so no change to the package json is required.

satish-a0: Okay

satish-a0 (approved):

csafreen (approved):

mwaiyee (approved):

PR #1356, opened 2025-12-09
Jerome ng/1087 add concept set delete

### LINKED ISSUE #1087: Add UI and backend to delete concept set

jerome-ng:
- Added delete button in concept set list
- owner can delete their own concept set
- concept set cannot be deleted if currently used in cohort def
- duplicate changes to concept set app
- Fixed dependencies (based on freshly cloned repo)

Screenshot 2025-12-09 at 10 17 01 AM
Screenshot 2025-12-09 at 10 16 50 AM
Screenshot 2025-12-09 at 7 28 29 PM


functions/terminology-svc/src/api/portal-api.ts
```diff
@@ -158,14 +158,18 @@ export class SystemPortalAPI {
...
+    datasetId: string
+  ): Promise<any> {
     console.info(`Portal request to delete concept set for id: ${id}`);
     const errorMessage = `Error while deleting concept set for id: ${id}`;
     try {
       const options = await this.createOptions();
       const url = `${
         this.url
-      }/user-artifact/concept_sets/${id}?datasetId=${encodeURIComponent(
+      }/user-artifact/${userId}/concept_sets/${id}?datasetId=${encodeURIComponent(
```
jerome-ng, line 172: using the endpoint with userId so only the owner can delete with this functionality

ui/apps/portal/package.json
```diff
@@ -20,6 +20,7 @@
     "@axa-fr/react-oidc": "6.10.9",
     "@babel/runtime": "^7.18.9",
     "@d4l/web-components-library": "^3.20.2",
+    "@data2evidence/d2e-starboard-wrap": "https://github.com/data2evidence/d2e-starboard-wrap.git#develop",
```
jerome-ng, line 23: many of these added dependencies are already being used in portal, but were dependencies of other apps that bun installed as part of the ur monorepo. With a fresh setup, there will be errors when not explicitly added as dependency. if the old yarn setup is still available in the system, i noticed sometimes it just works without explicitly adding

ui/apps/portal/src/components/Header/__tests__/Header.test.tsx
```diff
@@ -1,7 +1,6 @@
 import React from "react";
 import { render } from "@testing-library/react";
 import { MemoryRouter } from "react-router-dom";
-import "@testing-library/jest-dom/extend-expect";
```
jerome-ng, line 4: updated the jest-dom library, and it does not require this import

hengxian-jiang (approved):

functions/terminology-svc/src/api/portal-api.ts
satish-a0: If the concept set is being referenced in a PA cohort definition, once the concept set is deleted. Would the cohort definition still load? If the behaviour would be unpredictable, then its better to check if its being referenced and return an error avoid deletion, saying its being used currently. Atleast for this release.

jerome-ng: It errors when the concept set is unavailable, hence added check for existing usage.

satish-a0: Perfect!

satish-a0 (approved):

PR #1441, opened 2025-12-16
[Integrate-574] Move out init functions to main function

### LINKED ISSUE #574: Move out functions init code to the /functions folder
Most of the services have init functions and they're part of the sub-folder in the respective function. Change the structure of init code such that they're part of `/functions` folder and not within the sub-folder of the respective function. By doing so, this avoids duplication of `eszip` files created for each function.

hengxian-jiang:
### Merge Checklist


suwarnoong (commented):
There are many duplicate codes in this case. Is there a way to share these files or reference via relative path?

hengxian-jiang:
@suwarnoong Yes some code duplication will be there even after i cleaned up some unused files for migrations only. For example, env files, knex/typeorm entity files etc.

As they are different functions now, i am not sure if we can let them share the files through relative path, one solution is to port over some files to _shared folder.

suwarnoong (approved):

PR #1505, opened 2025-12-24
Jerome ng/1355 cdm config missing error

jerome-ng:
- Add error message for dataset permission check
- Fix dataset permission check

No errors:
Screenshot 2025-12-24 at 11 55 14 AM

Added permissions error message (not really needed, but leaving it in in case a permission issue occurs):
Screenshot 2025-12-24 at 11 55 23 AM

After permission check fix (backend):
Screenshot 2025-12-24 at 2 47 31 PM


functions/cdw-svc/src/main.ts
```diff
@@ -290,6 +291,22 @@ const initRoutes = (
...
+
+      // Determine if user has end-user privilege (system admin access)
+      let hasEndUserPrivilege = false;
+      try {
+        const userMgmtUrl = env.SERVICE_ROUTES?.usermgmt;
+        if (userMgmtUrl && token && user) {
+          const userMgmtApi = new UserMgmtAPI(userMgmtUrl);
+          const userId = user.getUser();
+          const userRoles = await userMgmtApi.getUserGroups(token, userId);
+          hasEndUserPrivilege = userRoles?.alp_role_system_admin === true;
```
jerome-ng, line 303: the main fix

satish-a0: NIT: This fix is a bit unconventional. On the other hand, this endpoint can be restricted only with an admin scope. Since researcher wouldnt need to access this endpoint. And the hasEndUserPrivilege can always be true.

satish-a0 (approved):

PR #1591, opened 2026-01-06
Fix data load and seeding

alicia-koh:
### Merge Checklist

Tested with data load
Screenshot 2026-01-06 at 11 27 32 AM
Screenshot 2026-01-06 at 11 24 33 AM
Screenshot 2026-01-06 at 11 15 00 AM


flows/base/package.json
```diff
@@ -673,18 +673,6 @@
                     "default": null,
                     "title": "Cache Schema Name"
                   },
-                  "result_schema": {
```
Zhimin-arya, line 676: So we don't use result schema ?

flows/base/omop_cdm_plugin/create.py
```diff
@@ -31,35 +31,30 @@ def create_datamodel_parent_task(cdm_version: str,
     '''
     logger = get_run_logger()
     tables_created = create_cdm_tables(schema_dao, cdm_schema, cdm_version, logger)
+    
     if tables_created:
-        create_concept_recommended_table(schema_dao, cdm_schema, logger)
```
csafreen, line 35: So we do not create `create_concept_recommended_table` if cdm schema and vocab schema are the same. why?

flows/base/omop_cdm_plugin/create.py
```diff
@@ -31,35 +31,30 @@ def create_datamodel_parent_task(cdm_version: str,
...
-            dbdao=schema_dao,
-            cdm_schema=cdm_schema,
-            vocab_schema=vocab_schema
-        )
-        
-    else:
-        # If creating schemas without vocab data
-        # Todo: Add insertion of cdm version to update flow
-        logger.info(f"Inserting dummy CDM Version '{cdm_version}'. Please update after loading vocabulary data.")
-        insert_cdm_version(
```
csafreen, line 56: And we do not insert into cdm_version when cdm schema and vocab schema is the same?

flows/base/package.json
alicia-koh: there's another parameter `results_schema` in the previous lines

flows/base/omop_cdm_plugin/create.py
alicia-koh: why does vocab schema need concept_recommended table? in the cache if it's in the cdm schema it will be copied anyway

flows/base/omop_cdm_plugin/create.py
alicia-koh: Yes in create_vocab_schema I'm setting schema name = vocab schema i.e. it's running the creation of tables on the vocab schema and we don't need to insert the cdm version there, only the cdm schema

flows/base/omop_cdm_plugin/create.py
csafreen: if its cdm_schema, and the schema doesn't have concept_recommended table then we used to create before, right?

csafreen: Don't think `concept_recommended` is part of standard cdm table and thats why we are creating it.

flows/base/omop_cdm_plugin/create.py
csafreen: But will be inserted when cdm_schema = vocab_schema but we are creating cdm tables?

flows/base/omop_cdm_plugin/create.py
alicia-koh: In this flow it creates three schemas. If it's the vocab schema it creates the omop tables. If it's the cdm schema it creates the omop tables + concept_recommended table. Then we create the results schema

flows/base/omop_cdm_plugin/create.py
alicia-koh: Yes for vocab schema we create omop tables there but don't insert cdm version. For cdm schema we create the omop tables + concept_recommended table + insert cdm version into cdm source table

csafreen (approved):

PR #1823, opened 2026-02-10
🐛 [PA] Not loading on Portal on 2nd render

khairul-syazwan:
### Merge Checklist

  **Summary**
   - Fixes Patient Analytics (Cohorts) plugin failing to render after navigating away and back.
   - Root cause: Vite-produced ES modules execute once per URL (browser module caching). The plugin relied on 
  re-execution; navigation produced a cached module that never re-ran its mount() logic.
   - This PR forces re-execution of Vue modules when context changes (cache-busting) and prevents orphaned script/style 
  tags by guarding async flows. SAP UI5 is intentionally kept as a singleton.

  **Root cause timeline**
   - 2026-01-26 — Vite migration (commit: 3ef5f6ac) changed build output from re-executable Webpack bundles to ES modules
   that execute once per URL.
   - 2026-01-29 — Single-SPA unload improvements (commit: 105cf3fd) made lifecycle management more consistent and exposed
   the ES module caching issue (it surfaced the bug; did not introduce it).

  **What changed**
   1. Force module re-execution for correct remounting
    - Add cache key (studyId + releaseId + timestamp) to module URLs so the browser treats each mount as a fresh module 
  (query param).
   2. Prevent orphaned scripts and memory leaks
    - Add isMounted guards in useEffect to abort async steps when the component unmounts; ensure cleanup callbacks only 
  run for scripts/styles added while mounted.


maggie-li-yd (commented):
Might it be better to use an explicit mount/unmount lifecycle? Cache busting and redownloading all JS bundles every time user navigates back to PA could be quite slow. The Jobs plugin `ui/apps/jobs/src/main.ts` seems to be using explicit mount/unmount lifecycle to solve this issue.

jerome-ng (approved):

PR #1880, opened 2026-02-20
Jerome ng/1758 materialize cohort e2e

jerome-ng:
Test was failing due to incorrect selectors. Materialize cohort also currently not working due to tables not created for demo cdm. the tests can be updated with `MATERIALIZATION_ENABLED = true` when it is fixed. for now it will at least run and test other aspects of the cohort builder.


scripts/e2e.sh
```diff
@@ -809,7 +809,7 @@ cmd_test() {
     mkdir -p "$temp_results" "$temp_ctrf"
 
     set +e
-    docker run --rm -it \
+    docker run --rm -i \
```
jerome-ng, line 812: removed `t` to allow automatic tools to use it better

mwaiyee (approved):

SantanM (approved):

PR #2488, opened 2026-05-14
Install sqlalchemy-hana based on deployment

alicia-koh:
### Merge Checklist


plugins/flows/base/package.json
```diff
@@ -124,6 +124,7 @@
         {
           "name": "create_cachedb_file_plugin",
           "entrypoint": "flows.create_cachedb_file_plugin.flow.create_cachedb_file_plugin",
+          "command": "bash -c 'uv pip install sqlalchemy-hana==2.2.0 && prefect flow-run execute'",
```
Zhimin-arya: `hana_load_plugin` also needs sqlalchemy-hana==2.2.0

plugins/flows/data_management/Dockerfile
```diff
@@ -90,6 +90,5 @@ WORKDIR /app
 
 # Place executables in the environment at the front of the path
 ENV PATH="/app/.venv/bin:$PATH"
-ENV INSTALL_SQLALCHEMY_HANA=true
```
Zhimin-arya: If it needs sqlalchemy-hana==2.2.0, should update package.json?

plugins/flows/base/package.json
csafreen: So we install when we do not need hana also? 
Can we have an if condition within the command and see if it works?

csafreen (changes requested):

plugins/flows/base/package.json
alicia-koh: Thanks I have added it

plugins/flows/data_management/Dockerfile
alicia-koh: Yes, instead of updating the dockerfile

plugins/flows/base/package.json
alicia-koh: > So we install when we do not need hana also?

Nope install only for the plugins that need it

> Can we have an if condition within the command and see if it works?

I think we will run into the same problem as before

Zhimin-arya (approved):

csafreen (approved):

p-hoffmann (approved):

PR #2496, opened 2026-05-17
add account linkning support

p-hoffmann:
### Merge Checklist


plugins/functions/alp-usermgmt/src/services/UserGroupService.ts
```diff
@@ -67,6 +73,7 @@ export class UserGroupService {
       },
       ...alpInfo
     }
+    this.maybeTriggerPhysionetSync(userId)
```
suwarnoong, line 76: This method is called inside `getUserGroupsMetadata` which is used when `USER_MGMT_ROLE_SOURCE=usermgmt`.

Our current default `USER_MGMT_ROLE_SOURCE=logto`. In this case, `maybeTriggerPhysionetSync` may not be called.

p-hoffmann: Fixed in getUserGroupsMetadataFromLogto: Added maybeTriggerPhysionetSync(user.id) before the return so the lazy reconcile also runs on the default USER_MGMT_ROLE_SOURCE=logto path. Gated on user since the sync needs the local user id.

suwarnoong (approved):

PR #2696, opened 2026-06-12
regression tests

SantanM:
Adds performance regression testing framework for D2E HTTP API endpoints.

This PR introduces a new performance regression test suite that measures response times against a locked baseline. It includes:

New regression test workflow (.github/workflows/_test-regression.yml) — sets up Docker services and runs performance tests
Performance test runner — TypeScript-based framework that parses HAR/curl files and compares response times against baseline
Initial baseline — pre-recorded p95 times for cohort definition and patient count endpoints
Two test scenarios — home page to cohort definition list flow (18 requests) and patient count/population bar chart endpoints
CI integration — integrated into the main docker-build-push workflow with a gate job (test_regression_gate)
Tests fail if any scenario exceeds 20% above baseline (configurable). Results are reported as tables with per-scenario deltas, and detailed reports are saved on failure.

image

### Merge Checklist

- [x] Documentation (Application guide, Admin guide, Markdown, Readme and/or Wiki)


hengxian-jiang (approved):

tests/regression/runner/httpClient.ts
```diff
@@ -0,0 +1,62 @@
...
+  for (let i = 0; i < config.repetitions; i++) {
+    const start = performance.now();
+    const res = await fetch(scenario.url, {
+      method: scenario.method,
+      headers,
+      body: scenario.body,
+    });
+    const elapsed = performance.now() - start;
+    // Drain the body so the connection is properly closed
+    await res.text();
```
suwarnoong, line 65: Should it check for `res.ok` or `res.status`?

.github/workflows/_test-regression.yml
```diff
@@ -0,0 +1,319 @@
...
+      reg_url:
+        required: true
+        type: string
+      prepull_flow_images:
+        required: true
+        type: string
+
+permissions:
+  contents: read
+  packages: write
```
suwarnoong, line 24: Do you need write permission?

tests/regression/runner/compare.ts
```diff
@@ -0,0 +1,47 @@
...
+      status: "no-baseline",
+      currentMinMs: result.minMs,
+      baselineMinMs: null,
+      deltaFraction: null,
+      minMs: result.minMs,
+      maxMs: result.maxMs,
+    };
+  }
+
+  const delta = (result.minMs - entry.minMs) / entry.minMs;
```
suwarnoong: NIT: possibility of division by zero?

suwarnoong: Should the delta based on minMs or p95 as per README?

SantanM: Not p95, it must be minMs. I will update the README.

.github/workflows/_test-regression.yml
SantanM: Yes, write permissions allow pulling of docker images from the `ghcr.io`. This is the same setting we've in other GH action running HTTP tests.

tests/regression/runner/compare.ts
SantanM: > NIT: possibility of division by zero?

`entry.minMs` is the baseline value of each test scenario -> 0 cannot be the right value as the "request-response" cycle will always be `x > 0`.

tests/regression/runner/httpClient.ts
SantanM: Not necessarily. These tests are meant for tracing and tracking the time taken by the backend to respond and not the response itself.

tests/regression/runner/httpClient.ts
```diff
@@ -0,0 +1,62 @@
...
+    samples.push(elapsed);
```
p-hoffmann, line 67: We never check the response status before recording the timing here. A fast `401`/`404`/`500` does no real DB work, so it lands well under the threshold and the scenario still passes even though the endpoint is broken. Can we assert `res.ok` (or the expected status) and fail the run otherwise? Otherwise the most likely outcome of this gate is a false green.

tests/regression/config.ts
```diff
@@ -0,0 +1,23 @@
+export const config = {
+  repetitions: Number(process.env.PERF_REPETITIONS ?? 3),
+
+  // Fraction above baseline p95 that triggers a warning (default 10%)
+  warnThreshold: Number(process.env.PERF_WARN_THRESHOLD ?? 0.1),
+
+  // Fraction above baseline p95 that fails the test (default 20%)
+  failThreshold: Number(process.env.PERF_FAIL_THRESHOLD ?? 0.2),
+
+  bearerToken: process.env.BEARER_TOKEN ?? "",
```
p-hoffmann, line 10: `bearerToken` defaults to `""`, and we only attach the `Authorization` header when it's truthy (`httpClient.ts:28`). So if `BEARER_TOKEN` isn't set — or the OIDC login step doesn't manage to scrape a token — every request goes out unauthenticated and 401s fast. Together with the missing status check, that's a green gate on broken auth. Can we fail fast when the token is empty?

tests/regression/runner/compare.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type Baseline = Record<string, { minMs: number }>;
+
+export function compareToBaseline(result: TimingResult, baseline: Baseline): CompareResult {
+  const entry = baseline[result.scenarioName];
+
+  if (!entry) {
+    return {
+      scenarioName: result.scenarioName,
+      status: "no-baseline",
```
p-hoffmann, line 25: `no-baseline` isn't `fail`, and the test only checks `status !== "fail"`, so any request without a baseline entry just slips through. The keys match today, but the moment the HAR gets re-captured and a request is added or dropped, the new one is never actually checked. I'd rather we treat `no-baseline` as a failure (or at least a hard warning).

tests/regression/runner/compare.ts
p-hoffmann: I don't think we should be gating on `minMs` alone — it's the least sensitive number we could pick, since a regression that pushes up p95/p99 but leaves the single fastest run untouched won't trip this. The README, scenario docs, and `config.ts` comments all describe this as a p95 gate anyway, so can we compare p95 (keeping min around as extra context is fine)?

p-hoffmann (changes requested):

p-hoffmann (approved):

PR #2708, opened 2026-06-15
add custom oidc connector

p-hoffmann:
### Merge Checklist


services/alp-logto/post-init/src/main.ts
```diff
@@ -480,19 +480,31 @@ button[name="submit"]{ background: #000080 !important; }`,
...
       signInMode: string;
+      socialSignIn: Object;
       signUp: Object;
       signIn: Object;
       socialSignInConnectorTargets: string[];
     } = {
       ...signinExperience,
-      signUp: { verify: false, password: true, identifiers: ["username"] },
+      signInMode: "SignInAndRegister",
+      socialSignIn: { automaticAccountLinking: true },
```
suwarnoong: Would this enable the "Register" button for all?

docker-compose-local.yml
```diff
@@ -33,6 +35,12 @@ services:
       # Bind-mounted source plugins win over bundled ones with the same package name.
       PLUGINS_DEV_PATH: "/usr/src/bundled-plugins:/usr/src/plugins"
       SQL_QUERY_TEMPLATES: '{"cohort-members": "SELECT {{COHORT_ID}} AS cohort_id, CURRENT_TIMESTAMP AS exported_at", "patient-count": "SELECT COUNT(*) AS patient_count FROM {{SCHEMA}}.person"}'
+      USERMGMT__AUTO_PROVISION_ENABLED: "true"
+      USERMGMT__AUTO_PROVISION_CONNECTORS: "physionet"
+      USERMGMT__AUTO_PROVISION_DEFAULT_TENANT_ID: "e0348e4d-2e17-43f2-a3c6-efd752d17c23"
+      USERMGMT__ENTITLEMENTS_SYNC_ENABLED: "true"
+      USERMGMT__ENTITLEMENTS_PHYSIONET_BASE_URL: "http://physionet-host:8000"
+      USERMGMT__ENTITLEMENTS_DATASET_MAPPING: '{"demo":"demoeicu/1.0"}'
```
suwarnoong: 1. Is this for your local or all developer will have to enable this by default?
2. These ENV does not seems to be all. Missing HOOK_URL, HOOK_SECRET?

services/alp-logto/post-init/src/main.ts
```diff
@@ -399,7 +399,7 @@ async function main() {
       `a[aria-label="Powered By Logto"] { display: none; }
 img[alt="app logo"] { height: 40px; margin-bottom: 20px; }
 button[name="submit"]{ background: #000080 !important; }`,
-    signInMode: "SignIn", //Disable user registration At Login screen
+    signInMode: process.env.LOGTO__CONNECTOR_CONFIG ? "SignInAndRegister" : "SignIn",
```
suwarnoong: `LOGTO__CONNECTOR_CONFIG` used by Entra & Entra External ID. These 2 connector does not need "Register" button

plugins/functions/alp-usermgmt/src/services/EntitlementsSyncService.ts
```diff
@@ -0,0 +1,245 @@
...
+    const url = `${baseUrl}/oauth/dataset-access/?slug=${encodeURIComponent(slug)}&version=${encodeURIComponent(version)}`
+    const controller = new AbortController()
+    const timer = setTimeout(() => controller.abort(), env.USERMGMT_ENTITLEMENTS_TIMEOUT_MS)
+    try {
+      const response = await fetch(url, {
+        headers: { Authorization: `Bearer ${physionetToken}` },
+        signal: controller.signal,
+      })
+      if (!response.ok) {
+        if (response.status === 404) return false
```
suwarnoong, line 185: is 404 assumed as "no-access"?

plugins/functions/alp-usermgmt/src/middlewares/add-user-object-to-req.ts
```diff
@@ -29,6 +29,13 @@ export const addUserObjToReq = async (req: IAppRequest, res: Response, next: Nex
     const sub = token[subProp]
     const idpUserId = oid! || sub!
 
+    // M2M tokens have sub === client_id; skip user lookup but still
+    // set a minimal req.user so downstream middleware doesn't crash.
+    if (sub === token.client_id) {
+      req.user = { userId: '', idpUserId: sub } as ITokenUser
```
suwarnoong: Maybe can consider to use a system-value for M2M userId like the one suggested by copilot `__service__`?

So that at https://github.com/OHDSI/Data2Evidence/pull/2708/changes#diff-32d5e0c48119e1fb6af4cf2a02b9279741ddfa8ba822d8944dab90435c78844aR30 can check for `__service__`

plugins/functions/alp-usermgmt/src/middlewares/grant-roles-by-scopes.ts
```diff
@@ -95,6 +151,11 @@ export const grantRolesByScopes = async (req: Request, res: Response, next: Next
       return next()
     }
 
+    const entitlementsSync = Container.get(EntitlementsSyncService)
+    await entitlementsSync.sync(userId!, sub, token).catch(err => {
+      logger.warn(`[Entitlements] sync threw for ${sub}: ${err}; keeping existing roles`)
+    })
```
suwarnoong, line 157: This would be called on every calls. Is this fine?

services/alp-logto/post-init/src/main.ts
p-hoffmann: It did. Registration is now opt-in behind a new `LOGTO__ENABLE_REGISTRATION` flag (default `false`), applied to both the base and social-connector sign-in experiences — so `SignInAndRegister` only kicks in when a deployment explicitly enables it. Fixed in d0848c05c.

services/alp-logto/post-init/src/main.ts
p-hoffmann: Agreed. Decoupled the Register button from `LOGTO__CONNECTOR_CONFIG` — it is now gated on the new `LOGTO__ENABLE_REGISTRATION` flag (default `false`), so Entra / Entra External ID keep a pure sign-in screen. Documented in `env-vars.md` and wired into `docker-compose.yml`. Fixed in d0848c05c.

plugins/functions/alp-usermgmt/src/middlewares/add-user-object-to-req.ts
p-hoffmann: Done. Added a `SERVICE_USER_ID = '__service__'` constant; M2M tokens (`sub === client_id`) are tagged with it instead of an empty string. `permitted-user-check` / `permitted-tenant-check` bypass only for `SERVICE_USER_ID`, and an unprovisioned end-user (empty `userId`) now gets a 403 instead of slipping through. Fixed in d0848c05c.

plugins/functions/alp-usermgmt/src/middlewares/grant-roles-by-scopes.ts
p-hoffmann: Gated and fail-soft: `sync()` returns immediately when `USERMGMT__ENTITLEMENTS_SYNC_ENABLED` is off, the base URL is unset, or the token has no PhysioNet claim — a no-op for non-PhysioNet deployments. When enabled it makes one upstream call per request to reconcile dataset access, with errors caught so existing roles are kept. We can add short-lived caching later if the per-request cost matters.

plugins/functions/alp-usermgmt/src/services/EntitlementsSyncService.ts
p-hoffmann: Yes, intentionally. A 404 from `/oauth/dataset-access/` means PhysioNet has no access record for this user/dataset (or the slug/version is unknown), so we treat it as "no access" and withhold/revoke the researcher role. Other non-OK statuses throw and are treated as transient, so existing roles are kept rather than revoked on a blip. Added a clarifying comment in d0848c05c.

docker-compose-local.yml
p-hoffmann: 1. Made it opt-in — both master switches (`USERMGMT__AUTO_PROVISION_ENABLED`, `USERMGMT__ENTITLEMENTS_SYNC_ENABLED`) now default to `false`, so the PhysioNet flow is off for local dev unless a dev sets them to `true` (the rest stays pre-wired as env-overridable defaults). Fixed in f7ae56786.
2. Intentional — the role hook is optional (`callRoleHook` returns early when the URL is unset), and local dev assigns roles via the entitlements sync, not the hook. The `HOOK_URL` / `HOOK_SECRET` vars are documented in `env-vars.md` for deployments that use a hook.

suwarnoong (approved):

Issue #2656, opened 2026-06-08
Add support for condition occurrence covariate concept sets in table1


p-hoffmann:
**Summary:**
Add support in `table1` for covariates derived from condition occurrence concept sets.

**Details:**
- Enable `table1` to accept condition occurrence concept sets as covariate definitions.
- Incorporate those concept set-based covariates into table1 generation and output.
- Ensure the implementation works consistently with existing covariate handling patterns.
- Update any relevant documentation describing table1 inputs and supported covariate types.

**Motivation:**
Users need to summarize cohorts using clinically meaningful condition-based covariates defined through concept sets. Supporting this directly in `table1` will make characterization workflows more flexible and easier to reproduce.

**Acceptance Criteria:**
- `table1` accepts condition occurrence concept sets as covariate inputs.
- Generated table1 output includes summaries for those covariates.
- Behavior is covered by tests.
- Documentation is updated if needed.

maksymtreshchov-d4l:
Checked on DEV, chrome latest, synpuf5 dataset

**QA Passed**

- [x] table1 accepts condition occurrence concept sets as covariate inputs.
- [x] Generated table1 output includes summaries for those covariates.
- [ ] Behavior is covered by tests.
- [ ] Documentation is updated if needed.

While the behavior is working for covariates as described. I couldn't find any new tests or documentation for behavior in related pull request, are they implemented in separate pull request, or will test coverage in acceptance criteria be fulfilled later? @amit-sharma-d4l 

Attachment:
Concept sets as covariate inputs

<img width="2557" height="1360" alt="Image" src="https://github.com/user-attachments/assets/279771ee-5655-4c04-a43b-c463aa656f12" />

Generated table1 output includes summaries for covariates

<img width="2557" height="1360" alt="Image" src="https://github.com/user-attachments/assets/c7ceab95-62f8-459c-a630-abb1d51c1c74" />

amit-sharma-d4l:
The core functionality is completed and PR is merged. Test cases and documentation are yet to be done, so moving this to 0.18 for tracking purposes.

PR #2724, opened 2026-06-17
Backend change for Concept Set Support in Table 1

amit-sharma-d4l:
Ticket - [https://github.com/OHDSI/Data2Evidence/issues/2656]

Changes - 
Added support for concept Ids in parquet-export endpoint. Now, an api call to this endpoint can include a list of concept ids - to be used for table1.

### DIFF
--- a/plugins/functions/parquet-export/index.ts
+++ b/plugins/functions/parquet-export/index.ts
@@ -81,6 +81,11 @@ function isValidWildcardFlag(val: string): boolean {
   return Number(val) === 1 || Number(val) === 0;
 }
 
+function isValidConceptIdArray(arr: unknown): arr is number[] {
+  if (!Array.isArray(arr)) return false;
+  return arr.every((item) => typeof item === "number" && Number.isInteger(item) && item > 0);
+}
+
 function sanitizeParamValue(value: string): string {
   return value.replace(/'/g, "''");
 }
@@ -107,6 +112,7 @@ const RESERVED_PLACEHOLDERS = new Set([
   "WILDCARD_FLAG3",
   "WILDCARD_FLAG4",
   "WILDCARD_FLAG5",
+  "CONCEPT_IDS",
 ]);
 
 function substituteTemplateParams(
@@ -214,6 +220,10 @@ function substituteTemplateParams(
     .replace(
       /\{\{WILDCARD_FLAG5\}\}/g,
       additionalParams["WILDCARD_FLAG5"] || "",
+    )
+    .replace(
+      /\{\{CONCEPT_IDS\}\}/g,
+      additionalParams["CONCEPT_IDS"] || "",
     );
 
   const remainingPlaceholders = extractPlaceholders(result);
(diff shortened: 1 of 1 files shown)


plugins/functions/parquet-export/index.ts
```diff
@@ -468,6 +478,14 @@ router.post("/", async (req: Request, res: Response) => {
       });
     }
 
+    const conceptIds = req.body.conceptIds as unknown | undefined;
+    if (conceptIds !== undefined && !isValidConceptIdArray(conceptIds)) {
```
p-hoffmann: An empty array slips through here — `isValidConceptIdArray([])` returns `true` because `[].every(...)` is `true` when there's nothing to check. But down at line 527 we only set `CONCEPT_IDS` when `length > 0`, so `{{CONCEPT_IDS}}` gets replaced with `""` and a template like `... IN ({{CONCEPT_IDS}})` becomes `IN ()` — a SQL syntax error that surfaces as an opaque 500. The `remainingPlaceholders` check won't save us either, since the placeholder is already gone and `CONCEPT_IDS` is reserved. Can we reject empty arrays here (or have the template tolerate an empty list)? Right now `conceptIds: []` is accepted but unusable.

plugins/functions/parquet-export/index.ts
```diff
@@ -81,6 +81,11 @@ function isValidWildcardFlag(val: string): boolean {
   return Number(val) === 1 || Number(val) === 0;
 }
 
+function isValidConceptIdArray(arr: unknown): arr is number[] {
+  if (!Array.isArray(arr)) return false;
+  return arr.every((item) => typeof item === "number" && Number.isInteger(item) && item > 0);
```
p-hoffmann: `item > 0` rejects concept_id `0`, which is a real OMOP value ("No matching concept") is excluding it intentional, or should it be `item >= 0`?

plugins/functions/parquet-export/index.ts
```diff
@@ -504,6 +523,11 @@ router.post("/", async (req: Request, res: Response) => {
       });
     }
 
+    // Handle conceptIds array
+    if (conceptIds && Array.isArray(conceptIds) && conceptIds.length > 0) {
+      additionalParams["CONCEPT_IDS"] = conceptIds.join(",");
+    }
```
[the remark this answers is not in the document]
p-hoffmann: This is valid and should be addressed. I traced it end-to-end: `CONCEPT_IDS` is now in `RESERVED_PLACEHOLDERS`, the `conditions` loop copies any reserved key in as `String(value)` with no validation, and unlike `CONCEPT_CODE2-5` the `{{CONCEPT_IDS}}` substitution is verbatim (`additionalParams["CONCEPT_IDS"] || ""`) — no `sanitizeParamValue`. So `{"conditions":[{"CONCEPT_IDS":"1) OR 1=1 --"}]}` with `conceptIds` omitted lands raw in the SQL, and the `;`-stripping doesn't help since no semicolon is needed.

It's latent today only because no SQL template references `{{CONCEPT_IDS}}` yet — it becomes a live injection the moment one does. Worth noting that `sanitizeParamValue` alone (single-quote escaping) wouldn't fully close it in a numeric context anyway, so the right fix is exactly as suggested: populate `CONCEPT_IDS` only from the validated `conceptIds` integer array and never from `conditions`.

p-hoffmann (approved):

PR #2732, opened 2026-06-18
[Analyze 2643] Refactor deep link handling to utilize sessionStorage for datasetId and query parameters

### LINKED ISSUE #2643: Page refresh from cohort builder leads to loading of first dataset instead of current dataset
**Description**
Users lose their place in the application on every browser refresh from cohort tab, forcing them to re-navigate to their dataset and tab. This is disruptive during normal workflows.

**Steps to Reproduce**
- Navigate to any dataset.
- Open the Cohorts tab within that dataset.
- Refresh the browser.
- Observe that the user is redirected to the first dataset in the server instead of remaining in current dataset -> Cohorts.

**Current Behavior**
When a user refreshes the browser while on the Cohorts tab, they are redirected to a different dataset instead of being returned to their previous view.

**Expected Behavior**
After a page refresh, the application should restore the user to the exact dataset and tab they were on — in this case, current dataset -> Cohorts.

**Acceptance Criteria**
- [ ] Refreshing the browser while on any dataset view restores the user to that same dataset and tab.
- [ ] The active dataset context is correctly reflected in the URL so that it survives a page reload.
- [ ] No redirect to a default or fallback dataset occurs unless the original dataset is no longer accessible.
- [ ] Behaviour is consistent across all datasets and tabs, not just reported issue.

https://github.com/user-attachments/assets/6e0fc32a-cde3-4b41-a15e-70f95c214944

jerome-ng:
### Merge Checklist

### DIFF
--- a/plugins/ui/apps/portal/src/hooks/useDeepLinkSync.ts
+++ b/plugins/ui/apps/portal/src/hooks/useDeepLinkSync.ts
@@ -6,10 +6,11 @@ import { Study } from "../types";
 import { config } from "../config";
 
 /**
- * Custom hook to sync dataset from URL on mount
+ * Custom hook to sync dataset from captured deep-link params on mount
  *
- * This hook runs once on component mount and checks for a `datasetId` URL parameter.
- * If found and valid, it auto-selects that dataset in Portal and navigates to the information page.
+ * Portal startup captures deep-link URL params in sessionStorage before this hook runs.
+ * If a stored datasetId is found and valid, this hook auto-selects that dataset in Portal
+ * and navigates to the captured target path or information page.
  *
  * @param datasets - List of available datasets user has access to
  * @param loading - Whether datasets are still loading
@@ -35,7 +36,6 @@ export const useDeepLinkSync = (datasets: Study[], loading: boolean) => {
     hasProcessedRef.current = true;
 
     syncDatasetFromUrl({
-      url: window.location.href,
       availableDatasets: datasets,
       setActiveDatasetId,
       setFeedback,
--- a/plugins/ui/apps/portal/src/utils/deepLinkHandler.ts
+++ b/plugins/ui/apps/portal/src/utils/deepLinkHandler.ts
@@ -46,8 +46,6 @@ export const isValidDatasetId = (datasetId: string | null | undefined): datasetI
 };
 
 export interface SyncDatasetFromUrlParams {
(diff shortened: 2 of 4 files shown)


plugins/ui/apps/portal/src/hooks/useDeepLinkSync.ts
```diff
@@ -34,11 +35,10 @@ export const useDeepLinkSync = (datasets: Study[], loading: boolean) => {
 
     hasProcessedRef.current = true;
 
-    syncDatasetFromUrl({
-      url: window.location.href,
```
jerome-ng, line 38: removed url from the deep link handler as it is not required. data is already stored in `index.tsx` so we only need to use the saved deep link params

plugins/ui/apps/portal/src/utils/deepLinkHandler.ts
```diff
@@ -70,33 +68,30 @@ export interface SyncDatasetFromUrlResult {
...
-    }
+export const syncDatasetFromUrl = ({
+  availableDatasets,
+  setActiveDatasetId,
+  setFeedback,
+  navigate,
+  basePath,
+}: SyncDatasetFromUrlParams): SyncDatasetFromUrlResult => {
+  const storedParams = loadDeepLinkParams();
+  const datasetId = storedParams?.datasetId ?? null;
```
jerome-ng, line 82: dataset id is retrieved from the deep link stored params , which is set on index.tsx

plugins/ui/apps/portal/src/utils/deepLinkHandler.ts
```diff
@@ -70,33 +68,30 @@ export interface SyncDatasetFromUrlResult {
...
-  setFeedback,
-  navigate,
-  basePath,
-}: SyncDatasetFromUrlParams): SyncDatasetFromUrlResult => {
-  const datasetId = extractDatasetIdFromUrl(url);
-  const storedParams = loadDeepLinkParams();
-
-  // No datasetId parameter — check if we have a path-based deep link (e.g., /researcher/wizards)
-  if (!isValidDatasetId(datasetId)) {
-    if (storedParams?.path && availableDatasets.length > 0) {
```
jerome-ng, line 89: This was making wizards and cohort builder use the first dataset id when a refresh is done. Instead we don't change the dataset id and let the default behaviour of portal change the dataset id using AppProvider

suwarnoong (approved):

PR #2736, opened 2026-06-19
Jerome ng/2072 cdm creation fails in manual test

### LINKED ISSUE #2072: CDM Creation fails in manual test
tested on develop.d2e.sg on synpuf5pct_cache.

https://docs.google.com/document/d/1obwn7FIEbVJvcLR4ePDJ5mSXCeTCHd0bYfcIW2gie0k/edit?tab=t.0

Step 11 fails with "Configuration invalid".

Manual tests possibly may not be expecting duckdb config

jerome-ng:
# Summary

Fixes several CDM config creation/editing failures found during manual testing. The changes focus on making CDW/Trex metadata calls safer, cleaning up opened connections, and avoiding cold-start attach races during config suggestions.

## What Changed

- Close analytics/Trex connections after CDM config metadata operations such as validation, autosave, activate, getColumns, and CDW suggestion services.
- Avoid opening an analytics connection for `attributeType_service`, which only needs config metadata.
- Update blank config defaults to use CDW service credentials instead of the old analytics service-tag lookup.
- Preserve suggestion endpoint response shape by returning `{ "data": [] }` for column/table suggestion failures.
- Add a per-destination attach guard for Trex connections so the first cold `getConnection()` is serialized, then later requests use the normal fast path.

## Why

Manual testing exposed multiple related failure modes:

- Repeated metadata calls could leave Trex connections open until Trex became unstable.
- `configDefaults` could fail with `Found 0 matching services` when looking up the old `analytics` tag.
- Some UI flows passed dataset/config context differently, especially after changing the dataset selector.
- Cold parallel suggestion requests could race while attaching DuckDB validation schemas and fail with `validation_schema already exists`.

## Validation

- Reproduced and diagnosed the CDW suggestion attach race.
- Reproduced the unclosed-connection failure with cleanup reverted: sequential requests timed out around request 815.
- Verified the cleanup fix with 5000 `column_suggestion_service` requests at concurrency 10 plus a successful follow-up request.
- Compared no-lock, always-mutex, and fast-path guard approaches. The fast-path guard completed 5000 / 5000 requests and stayed effectively at baseline latency, while the always-mutex version was materially slower.
- Confirmed no `validation_schema already exists`, `Failed to attach database`, `UncaughtException`, `event loop error`, or request timeout errors during the final benchmark run.
- Successfully completed the `CDM creation` manual test locally.


plugins/functions/cdw-svc/src/main.ts
```diff
@@ -31,6 +31,25 @@ const noCache = (req, res, next) => {
   next();
 };
 
+const sendCDWServiceError = (res, action) => {
+  if (
+    action === "column_suggestion_service" ||
```
jerome-ng, line 36: Preserving existing behaviour

plugins/functions/cdw-svc/src/main.ts
```diff
@@ -275,50 +294,69 @@ const initRoutes = (
...
-        ),
-        isTestEnvironment
-      ).invokeService(req.query.action, req.body, hasEndUserPrivilege, (err, result) => {
-        if (err) {
-          log.error(err);
-          return res.status(500).send({});
+          isTestEnvironment,
+        );
+
+        if (req.query.action === "attributeType_service") {
```
jerome-ng, line 312: Early return as this action does not need analytics service

brandantck (approved):

plugins/functions/cdw-svc/src/utils/trexConnection.ts
```diff
@@ -34,6 +37,113 @@ const parseSql = (
...
+    databaseCode,
+    schemaName,
+    vocabSchemaName,
+    resultsSchemaName,
+    {
+      duckdb: parseSql,
+    }
+  );
+
+const getConnectionWithAttachGuard = async (
```
brandantck, line 83: NIT: Can add some comments to explain why this attach guard is required?

jerome-ng: added 
```
  // Trex lazily attaches DuckDB aliases during the first connection for a
  // destination. Cold parallel requests can race that attach and fail with
  // "validation_schema already exists", so serialize the first successful open
  // per destination in this worker. Once ready, skip the lock so later opens
  // for the same destination can run in parallel on the normal fast path.
  ```

PR #2766, opened 2026-06-24
Improve logging in DC and DQD

csafreen:
### Merge Checklist


plugins/flows/base/dqd_plugin/utils.py
```diff
@@ -0,0 +1,19 @@
+from _shared_flow_utils.diagnostics import collect_bounded_diagnostics
+
+
+def collect_dqd_diagnostics(
+    output_folder: str,
+    output_file: str | None = None,
+    max_error_files: int = 10,
+    max_log_files: int = 3,
+    max_chars_per_file: int = 4000,
```
alicia-koh, line 9: collect_bounded_diagnostics also defaults `max_chars_per_file` to 4000, can choose to set it in either function

plugins/flows/base/data_characterization_plugin/utils.py
```diff
@@ -15,6 +16,30 @@ def get_failed_analysis_ids(output_folder: str) -> list[int] | None:
...
+        return int(file_path.stem.split("_")[-1])
+    except ValueError:
+        return float("inf")
+
+
+def collect_achilles_diagnostics(
+    output_folder: str,
+    max_error_files: int = 10,
+    max_log_files: int = 3,
+    max_chars_per_file: int = 4000,
```
alicia-koh, line 30: collect_bounded_diagnostics also defaults `max_chars_per_file` to 4000, can choose to set it in either function

csafreen: There is no benefit to do this change right now as its just setting the default here. The functions calling this has the option to pass custom value to`max_chars_per_file`

plugins/flows/base/dqd_plugin/utils.py
csafreen: There is no benefit to do this change right now as its just setting the default here. The functions calling this has the option to pass custom value to`max_chars_per_file`

alicia-koh (approved):

plugins/flows/_shared_flow_utils/diagnostics.py
```diff
@@ -0,0 +1,54 @@
+from pathlib import Path
+from typing import Callable
+
+
+def _read_text_snippet(file_path: Path, max_chars: int) -> str:
+    try:
+        text = file_path.read_text(errors="replace")
```
suwarnoong, line 7: This would load the entire file into memory first, and truncate at later step up to max chars. For large diagnostic file, it would be a concern

PR #2767, opened 2026-06-24
Dropdown is not opening for Patient Cohort filter card

khairul-syazwan:
Resolves https://github.com/data2evidence/project/issues/14

Screen Recording 2026-06-24 at 11 13 08 AM

Scope is not lost.
Screenshot 2026-06-24 at 3 13 33 PM
Screenshot 2026-06-24 at 3 13 38 PM
Screenshot 2026-06-24 at 3 13 43 PM

## Context
On HANA datasets, the cohort definition ID filter card showed a numeric range input instead of a searchable tag input, so the frontend never called the `analytics-svc/api/services/values` endpoint. Once that was corrected, two additional backend bugs surfaced that prevented results from being returned.
## Summary
Three root causes were found and fixed in sequence:
### 1. Wrong attribute type in HANA CDW seed config
`plugins/functions/mri-pg-config/src/db/seeds/03_Config.ts`
`cohortdefinitionid` in `cdwConfig` (the main HANA CDW config, used by the `OMOP_HANA` PA config) had `"type": "num"`. The frontend `Constraint.vue` only renders a searchable tag input — which triggers the values API — when `model.props.type === 'text'`. With `"num"`, it rendered a numeric range slider instead, so no API call was ever made.
Fix: Changed `"type"` to `"text"` and added the missing HANA-specific reference fields (`referenceFilter` using `CONTAINS ... FUZZY (0.5)`, `referenceExpression`, `includeDescendants`, `includeDescendantsExpression`) to match the pattern already used in `omopHanaLeanCdwConfig`.
### 2. Missing referenceExpression caused a backend crash
`plugins/functions/query-gen-svc/src/svc/DomainValuesSvc.ts`
`DomainValuesSvc` sets `exprToUse = "referenceExpression"` when `useRefValue: true` (which the HANA config has). Without the `referenceExpression` field, the constructor set `attrRefExpression = undefined`, and `getDistinctValuesFromReference` then crashed with `TypeError: Cannot read properties of undefined (reading 'match')` at line 453.
This was resolved by fix #1 (adding `referenceExpression` to the config).
### 3. Empty search returned no results
`plugins/functions/query-gen-svc/src/svc/DomainValuesSvc.ts`
On initial dropdown open (empty search), the values API returned `{"data":[]}`. `getDistinctValuesFromReference` replaced `@SEARCH_QUERY` with the internal placeholder `__EMPTY_SEARCH__` and built a standalone `WHERE CONTAINS (..., '%__EMPTY_SEARCH__%', FUZZY (0.5))` clause. The existing `removeEmptySearchCondition` helper only removed `AND` clauses, not a standalone `WHERE` clause, so it returned the SQL unchanged. HANA's `CONTAINS` found no rows matching the literal string `__EMPTY_SEARCH__`.
An initial attempt fixed this by guarding `aliasedRefFilter` so it was only built when `searchQuery` was non-empty. However, reviewer feedback noted this would also drop important scoping predicates for other reference-backed filters (e.g. `DOMAIN_ID = 'Gender' AND STANDARD_CONCEPT = 'S'` for Gender), causing them to return unconcepted results.
Fix: Replaced the guard with an improved `removeEmptySearchCondition` that now handles both:
- `AND`-joined empty-search predicates — removes only that predicate, preserving scoping.
- Lone / leading `WHERE` empty-search predicates — removes the predicate, or the entire `WHERE` clause if it's the only predicate.
This fixes the Patient Cohort Cohort Definition ID dropdown on HANA while keeping scoping intact for filters like Gender, Race, and Ethnicity on both Postgres and HANA.
### Files changed
- `plugins/functions/mri-pg-config/src/db/seeds/03_Config.ts`
- `plugins/functions/query-gen-svc/src/svc/DomainValuesSvc.ts`
- `plugins/functions/query-gen-svc/spec/svc/DomainValuesSvc_test.ts`
### Merge Checklist
- [x] Updated Manual tests / Demo Config
- [x] following best practices in [code review doc](../code_review.md)


plugins/functions/mri-pg-config/src/db/seeds/03_Config.ts
maggie-li-yd, line 1: If I'm not wrong, the seed update in `03_Config.ts` only works on fresh installs. We might still need to manually update the config in the mt sinai set up.

plugins/functions/query-gen-svc/src/svc/DomainValuesSvc.ts
```diff
@@ -471,7 +471,11 @@ function getDistinctValuesFromReference(
         placeholderAliasMap,
         attrRefExpression
     );
-    const aliasedRefFilter = attrRefExpression
+    // Skip the reference filter for empty searches so all values are returned.
+    // removeEmptySearchCondition() only strips AND clauses; when the filter is
+    // the sole WHERE clause it leaves __EMPTY_SEARCH__ in the SQL, which HANA's
+    // CONTAINS (and Postgres JARO_SIMILARITY) won't match.
+    const aliasedRefFilter = (attrRefExpression && referenceFilter && searchQuery)
```
maggie-li-yd: Could this make the empty search result list for other filters, like Gender, lose scoping?

plugins/functions/mri-pg-config/src/db/seeds/03_Config.ts
khairul-syazwan: Yes, this is just to align it with postgres and hana lean config seeds. Manual updates should not be required since they are using the lean config.

plugins/functions/query-gen-svc/src/svc/DomainValuesSvc.ts
khairul-syazwan: Good catch. I've made changes to the code, and tested that it works for both Postgres and HANA

maggie-li-yd (approved):

PR #2824, opened 2026-07-01
revert installation of sqlalchemy-hana during flow run

alicia-koh:
### Merge Checklist


docker-compose.yml
```diff
@@ -363,6 +363,7 @@ services:
       D2E_SWAP_LIMIT: ${D2E_SWAP_LIMIT:-}
       D2E_DUCKDB_MEMORY_LIMIT: ${D2E_DUCKDB_MEMORY_LIMIT:-15GB}
       D2E_DUCKDB_THREADS: ${D2E_DUCKDB_THREADS:-}
+      INSTALL_SQLALCHEMY: ${INSTALL_SQLALCHEMY:-}
```
Zhimin-arya, line 366: should we set `bash -c 'uv pip install sqlalchemy-hana==2.2.0 && prefect flow-run execute'` as default value of `INSTALL_SQLALCHEMY`

Zhimin-arya (commented):
Just one comment on `INSTALL_SQLALCHEMY` default value, the rest looks fine to me.

docker-compose.yml
alicia-koh: no because that will require internet access to download sqlalchemy-hana when flow run starts and we don't want that

Zhimin-arya (approved):

docker-compose.yml
suwarnoong, line 366: INSTALL_SQLALCHEMY in the d-c.yml but INSTALL_SQLALCHEMY_HANA in Dockerfile. Do I miss something?

alicia-koh: The dockerfile `INSTALL_SQLALCHEMY_HANA` is an env we set so when flow run container starts it runs a custom command to install sqlalchemy hana first. `INSTALL_SQLALCHEMY` is a variable that contains that custom command which we set in .env and import into the prefect workpool. The workpool will spin up containers with the command we set.

suwarnoong: cool, thanks for explaining

suwarnoong (approved):

docker-compose.yml
SantanM: @alicia-koh - The naming of the env variable does not sound right when it is intended for providing custom command for starting the container. Could you please change it to something like - `PREFECT__FLOWRUN_STARTUP_CUSTOM_CMD`?

p-hoffmann (approved):

Issue #2843, opened 2026-07-03
Implement d2e wizards for non-hana datasets


csafreen:
## Summary
Integrate d2e wizards for non-hana datasets

## Problem
For non-hana datasets, the concept name is being passed to the parquet export endpoint in the request body when it is expecting a concept code. The sql query therefore fails to be executed because it uses the wrong column for mapping.

<img width="2498" height="1258" alt="Image" src="https://github.com/user-attachments/assets/9bb5b81e-0885-41fc-ae31-01845706579b" />

<img width="1024" height="896" alt="Image" src="https://github.com/user-attachments/assets/b8df2c17-f9e9-4711-b46b-792003ff6bf5" />

## Goals and scope
- Wizards config:
  - For non-hana datasets, the `concept_id` of the condition should be passed when a condition is selected
  - For hana datasets, the `concept_code` of the condition should be passed when a condition is selected
- Parquet export endpoint:
  - It should be updated to expect either a `concept_id` or `concept_code` from the request body based on the dialect passed
- SQL Templates:
  - For non-hana sql templates, it should be updated to use `concept_id`
- Python dashboard code:
  - For non-hana dashboards, it should be updated to use `concept_id`

## Acceptance Criteria
- Incidence rate, prevalence rate, mortality rate, and demographics dashboards are visible for synpuf5pct dataset in the D2E develop portal
- Incidence rate, prevalence rate, mortality rate, and demographics dashboards are visible for HANA_CDMSYNPUF1K dataset in the D2E develop portal

seolhwa-kim-data4life:
@p-hoffmann i can not still assign this to myself. but will start to work on this one.

PR #2889, opened 2026-07-13
Issue-2843/compatible wizard [help needed]

seolhwa-kim-data4life:
# Description #2843 

 When a condition is selected in the Wizards dashboards, the concept name (e.g. "Acute bronchitis") was being sent to the endpoint as CONCEPT_CODE, but the SQL joins on concept_code, so it never matched.
Per the ticket, non-HANA (duckdb/postgres) datasets should send the concept_id and HANA datasets should send the concept_code.

Here is following changes in wizard-dashboard (https://github.com/data2evidence/wizards-dashboard/pull/24)

# Verified locally

1. Incidence rate, prevalence rate, mortality rate, and demographics dashboards are visible for synpuf5pct dataset in the D2E develop portal

image

# Help needed

I wish a careful review / helping me test on these — this isn't my main area (nor ts / vue / wizards test).
How can i also check Incidence rate, prevalence rate, mortality rate, and demographics dashboards are visible for HANA_CDMSYNPUF1K dataset?


jerome-ng:
Thanks for your PR @seolhwa-kim-data4life. I see changes have been merged in https://github.com/data2evidence/wizards-dashboard/pull/24 and this PR focuses on

1. Providing the dialect to the shiny app
2. Modifying the configs for the data to be available (concept id) in this case

No 2 is less trivial as the suggested modifications will also change how cohort builder behaves. We will discuss this during the analyze workshop to figure out what should be used and create appropriate issues for both these points.

jerome-ng:
@seolhwa-kim-data4life Hey. Are you able to change the use of concept id to use concept codes instead? The issue is that in order to use concept id, it must be available in the filter cards as an attribute, but we do not want to show the concept id. We can use the SNOMED ones which should be available in the demo dataset. we possibly might not need the dialect anymore if we use the code

seolhwa-kim-data4life:
> @seolhwa-kim-data4life Hey. Are you able to change the use of concept id to use concept codes instead? The issue is that in order to use concept id, it must be available in the filter cards as an attribute, but we do not want to show the concept id. We can use the SNOMED ones which should be available in the demo dataset. we possibly might not need the dialect anymore if we use the code

@jerome-ng now returns the SNOMED concept_code 
One open question on point 2: the HANA configs already expose condition_concept_code with  filtercard.visible: true, but I've kept the new duckdb entry hidden (visible: false, category: false). Should duckdb match HANA and show it, or stay hidden? 

https://github.com/data2evidence/wizards-dashboard/pull/27

here it is now using concept_code back in the pr. idk what to do with  vocabulary_id = 'SNOMED'  tho. maybe will see with alicia.

PR #2983, opened 2026-07-27
feat(cohort_discovery): Hutch Bunny Relay poller flow plugin group

ohdsi-trex:
## Summary

Adds a new `cohort_discovery` d2e flow plugin group: a scheduled, short-lived Prefect flow that acts as a federated cohort-discovery node. On each tick it polls a Hutch **Relay Task API** via Hutch Bunny's native `TaskApiClient`/`PollingService`, pulls **at most one** native RQuest task, executes it against the dataset's OMOP database through Bunny's own DB client, submits results back to the Relay, and persists a normalized Prefect artifact.

**Scope: flow + shared-worker provisioning only.** No `jobplugins` API endpoint and no Jobs-page UI (schedule controls, run history/status, results rendering) — consumption is the persisted Prefect artifact plus Prefect's own run views. A UI/API layer may be planned separately.

### Architecture

Hutch Bunny requires Python >= 3.13, while d2e's flow stack and DBDao are on 3.12. Rather than co-resolving the two dependency stacks, they are **isolated** following the `data_transformation` `ner` pattern:

- **`default` env (3.12)** — runs the Prefect `@flow`: resolves the dataset's DB/cachedb credentials via **DBDao**, maps them to Bunny `DATASOURCE_*` config, invokes the child, parses its JSON, and writes the artifact.
- **`bunny` env (3.13, `no-default-feature = true`)** — self-contained; runs `bunny_runner.py` (`poll_for_tasks(max_iterations=1)` → `execute_query` → `send_results`).

Bunny is a **git dependency (not on PyPI)**, pinned to the exact v1.7.0 commit `a4121dc7a37660638763ed20b11d8882a7ab8edd`. A `pixi.lock` covering both environments is committed, since runtime provisioning is `--frozen`.

Failures are **hard-fails**, never silent or approximate: unsupported DB dialects (Postgres + DuckDB cachedb only), child execution/config errors, and unsupported RQuest constructs all fail the run and surface a structured error.

### Design docs

Included in this PR as the decision record:

- `trex/specs/2026-07-24-cohort-discovery-design.md` — approved design spec (architecture, interfaces, artifact envelope, error handling).
- `trex/plans/2026-07-24-cohort-discovery-bunny.md` — task-by-task implementation plan this PR executes.
- `trex/plans/2026-07-24-cohort-discovery-compat-gate.md` — the pre-implementation compatibility gate, executed and **passed** empirically (both envs build and coexist; pinned Bunny API surface recorded).

### Changed files

New group `plugins/flows/cohort_discovery/` — `flow.py` (parent), `bunny_runner.py` (child), `bunny_config.py` (DBDao → `DATASOURCE_*`), `types.py`, `pyproject.toml`, `pixi.lock`, `package.json` (`trex.flow` manifest), `setup_assets.sh`, `README.md`, `BUNNY_NOTES.md`, and tests. Plus a 4-line additive `bunny` branch in `services/alp-dataflow-gen-worker/provision-envs.sh`, mirroring the existing `ner` block.

## Test Plan

- [x] `pytest cohort_discovery_plugin/tests -m "not integration"` → **9 passed** (models, credential mapping incl. unsupported-dialect hard-fail, child runner with mocked Bunny, parent flow artifact construction + hard-fail path)
- [x] Both Pixi envs resolve: `pixi install` (3.12.13) and `pixi install -e bunny` (3.13.14, Bunny from the pinned commit)
- [x] Bunny API surface verified against the pin; import-time env ordering confirmed (child must populate env before importing Bunny — it does)
- [ ] **End-to-end against a live Relay + real OMOP dataset** — requires Relay credentials and a dataset fixture supplied at deployment. An env-gated integration test (`-m integration`, skips unless `TEST_OMOP_DUCKDB` + `TEST_RELAY_URL` are set) plus a stub-relay/fixture guide are included in `cohort_discovery_plugin/tests/README.md`.

### Deployment notes

Required deploy-time env is documented in `plugins/flows/cohort_discovery/README.md`: `TASK_API_BASE_URL` (incl. `/link_connector_api`), `TASK_API_USERNAME`/`PASSWORD`/`TYPE`, `TASK_API_ENFORCE_HTTPS`, `COLLECTION_ID` (= `datasetId`), `LOW_NUMBER_SUPPRESSION_THRESHOLD`, `ROUNDING_TARGET`, and `COHORT_DISCOVERY_ICD_MAIN_ENABLED` (default off). Cadence is owned by the Prefect deployment schedule; one poller deployment per dataset.


plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
```diff
@@ -0,0 +1,86 @@
...
+    `DaoBase.create_ibis_connection_url`, which builds `duckdb://{databaseName}`),
+    so we expose it to `build_bunny_env` as `duckdb_path` via a small shim.
+    """
+    from types import SimpleNamespace
+
+    from _shared_flow_utils.dao.DBDao import DBDao  # lazy: DBDao pulls in ibis, keep off module import path
+
+    dao = DBDao(database_code=options.databaseCode, cache_id=options.cacheId or options.databaseCode)
+    creds = dao.tenant_configs
+    if str(getattr(creds, "dialect", "")).lower() == "duckdb":
```
csafreen: duckdb is not a valid dialect in d2e. its trex.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
```diff
@@ -0,0 +1,86 @@
...
+                  "generatedAt": datetime.now(timezone.utc).isoformat()},
+    )
+
+@flow(log_prints=True)
+def cohort_discovery_plugin(options: CohortDiscoveryOptions) -> ArtifactEnvelope:
+    logger = get_run_logger()
+    logger.info(f"cohort_discovery start: dataset={options.datasetId}")
+
+    creds, schema = _resolve_credentials(options)
+    base_env = {k: os.environ[k] for k in (
```
csafreen, line 77: Where does these environment variable values come from? alp-dataflow-gen-worker? If yes, then need an update in function/package.json

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_config.py
```diff
@@ -0,0 +1,28 @@
+from typing import Any
+
+_PG_DIALECTS = {"postgres", "postgresql"}
+_DUCKDB_DIALECTS = {"duckdb"}
```
csafreen: Confirm if its trex in data2evidence repo

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_config.py
```diff
@@ -0,0 +1,28 @@
...
+        env["DATASOURCE_DB_HOST"] = str(creds.host)
+        env["DATASOURCE_DB_PORT"] = str(creds.port)
+        env["DATASOURCE_DB_DATABASE"] = str(creds.databaseName)
+        env["DATASOURCE_DB_SCHEMA"] = schema
+        env["DATASOURCE_DB_USERNAME"] = str(creds.user)
+        env["DATASOURCE_DB_PASSWORD"] = _secret(creds.password)
+    elif dialect in _DUCKDB_DIALECTS:
+        env["DATASOURCE_DB_DRIVERNAME"] = "duckdb"
+        # Attribute per BUNNY_NOTES.md / CacheDBCredentialsType (Task 1).
+        env["DATASOURCE_DUCKDB_PATH_TO_DB"] = str(creds.duckdb_path)
```
csafreen: Handle the connection through TrexDao instead of directly connecting with duckdb file

plugins/flows/cohort_discovery/cohort_discovery_plugin/types.py
```diff
@@ -0,0 +1,24 @@
...
+class ChildTaskResult(BaseModel):
+    # One resolved RQuest result, as emitted by the child (from RquestResult.to_dict()).
+    analysis: Optional[str] = None      # None => availability; else distribution code
+    count: Optional[int] = None
+    distributions: dict[str, Any] = {}
+    raw: dict[str, Any] = {}            # full to_dict() for traceability
+
+class ChildResult(BaseModel):
+    results: list[ChildTaskResult] = []
+    error: Optional[str] = None
```
csafreen, line 32: Handle the comment added by copilot

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_runner.py
```diff
@@ -0,0 +1,44 @@
...
+        d = result.to_dict()
+        collected.append({
+            "analysis": task_data.get("analysis"),
+            "count": d.get("count"),
+            "distributions": d.get("distributions", {}),
+            "raw": d,
+        })
+
+    try:
+        PollingService(client, handler, settings).poll_for_tasks(max_iterations=1)
```
csafreen, line 42: Do not hardcode max_iterations. Keep it as an input to the prefect flow

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
```diff
@@ -0,0 +1,86 @@
...
+from .types import CohortDiscoveryOptions, ChildResult, ArtifactEnvelope
+from .bunny_config import build_bunny_env
+
+os.environ["plugin_name"] = "cohort_discovery_plugin"
+
+def _resolve_credentials(options: CohortDiscoveryOptions):
+    """Resolve dataset DB/cachedb credentials via DBDao (3.12 parent). Returns (creds, schema).
+
+    Real-attribute reconciliation (verified against
+    `_shared_flow_utils/types.py::DBCredentialsType`): Postgres creds already
```
csafreen: Remove this comment

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
```diff
@@ -0,0 +1,86 @@
...
+    base_env = {k: os.environ[k] for k in (
+        "TASK_API_BASE_URL", "TASK_API_USERNAME", "TASK_API_PASSWORD", "TASK_API_TYPE",
+        "LOW_NUMBER_SUPPRESSION_THRESHOLD", "ROUNDING_TARGET",
+    ) if k in os.environ}
+    env = build_bunny_env(creds, schema=schema, collection_id=options.datasetId, base_env=base_env)
+
+    child = ChildResult.model_validate_json(_run_child(env))
+    if child.error:
+        raise RuntimeError(f"cohort_discovery hard-fail: {child.error}")
+
```
csafreen, line 94: Handle the comment by copilot on lines R75 to R77

plugins/flows/cohort_discovery/README.md
```diff
@@ -0,0 +1,78 @@
+# cohort_discovery flow group
+
+A scheduled, short-run Prefect flow that connects a d2e dataset to a **Hutch
+Relay Task API** through [Hutch Bunny](https://github.com/Health-Informatics-UoN/hutch-bunny).
+
+On each tick the flow:
```
csafreen: By tick do you every trigger of the flow?

plugins/flows/cohort_discovery/README.md
```diff
@@ -0,0 +1,78 @@
...
+| `TASK_API_USERNAME`                 | Relay Task API username.                                                                   |
+| `TASK_API_PASSWORD`                 | Relay Task API password.                                                                   |
+| `TASK_API_TYPE`                     | `a` = availability, `b` = distribution. (Bunny has no `c`.)                                |
+| `COLLECTION_ID`                     | The Relay collection id — **equal to the d2e `datasetId`** this poller serves.             |
+| `LOW_NUMBER_SUPPRESSION_THRESHOLD`  | Low-number suppression threshold applied to results (e.g. `10`).                           |
+| `ROUNDING_TARGET`                   | Rounding target applied to results (e.g. `10`).                                            |
+| `COHORT_DISCOVERY_ICD_MAIN_ENABLED` | Enable ICD_MAIN distribution tasks. **Default off.** When off, ICD_MAIN distribution tasks are excluded. |
+| `TASK_API_ENFORCE_HTTPS`            | Set **`false`** for a non-HTTPS Relay `TASK_API_BASE_URL`. Bunny's `DaemonSettings` enforces HTTPS by default. |
+
```
csafreen, line 45: Handle comment by copilot on lines R51 to R52

csafreen (commented):
Take a look at the comments and respond and make changes if neede

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_config.py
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1. The DuckDB branch is gone entirely — cachedb now routes through Trex's pgwire endpoint (see the thread on `build_bunny_env`), and `DATASOURCE_DB_SCHEMA` is set in **every** supported branch. There is also an explicit guard that raises with a clear message when the schema is empty, plus a `_REQUIRED` completeness check, since `DaemonSettings` declares `DATASOURCE_DB_SCHEMA` with no default and would otherwise fail with an opaque error at Bunny import time.

plugins/flows/cohort_discovery/cohort_discovery_plugin/tests/test_bunny_config.py
```diff
@@ -0,0 +1,33 @@
...
+    assert env["DATASOURCE_DB_SCHEMA"] == "cdm"
+    assert env["DATASOURCE_DB_USERNAME"] == "u"
+    assert env["DATASOURCE_DB_PASSWORD"] == "pw"
+    assert env["COLLECTION_ID"] == "ds1"
+
+def test_duckdb_env_mapping():
+    creds = SimpleNamespace(dialect="duckdb", duckdb_path="/data/ds1.duckdb")
+    env = build_bunny_env(creds, schema="cdm", collection_id="ds1", base_env={})
+    assert env["DATASOURCE_DB_DRIVERNAME"] == "duckdb"
+    assert env["DATASOURCE_DUCKDB_PATH_TO_DB"] == "/data/ds1.duckdb"
```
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1. The DuckDB test is replaced by a Trex test that asserts `DATASOURCE_DB_SCHEMA` is set, and the Postgres test now asserts it too. Added `test_missing_schema_hard_fails` to pin the guard directly.

plugins/flows/cohort_discovery/cohort_discovery_plugin/tests/test_integration.py
```diff
@@ -0,0 +1,25 @@
...
+        pytest.skip("Set TEST_OMOP_DUCKDB and TEST_RELAY_URL to run this test")
+    env = {
+        **os.environ,
+        "DATASOURCE_DB_DRIVERNAME": "duckdb",
+        "DATASOURCE_DUCKDB_PATH_TO_DB": str(duckdb_file),
+        "COLLECTION_ID": "ds-test",
+        "TASK_API_BASE_URL": os.environ["TEST_RELAY_URL"],
+        "TASK_API_USERNAME": "u", "TASK_API_PASSWORD": "p",
+        "LOW_NUMBER_SUPPRESSION_THRESHOLD": "10", "ROUNDING_TARGET": "10",
+    }
```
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1 — both `DATASOURCE_DB_SCHEMA` and `TASK_API_ENFORCE_HTTPS=false` are now set in the subprocess env. Also added a comment noting this test drives Bunny's own DuckDB client as a cheap local fixture and is **not** the production path (production is Trex pgwire).

plugins/flows/cohort_discovery/cohort_discovery_plugin/types.py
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1 — all mutable defaults now use `Field(default_factory=...)`, matching `dataflow_ui_plugin/types.py`. Added a test asserting instances don't share default state.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_runner.py
```diff
@@ -0,0 +1,44 @@
+"""Runs inside the isolated Bunny 3.13 pixi env. Config comes from env (set by parent)."""
+import json, sys, traceback
+
+def _build_results_modifiers(settings):
```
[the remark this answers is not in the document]
ohdsi-trex: Good catch, but the fix went the other way — the toggle was **removed** rather than enforced (d3aabe91).

Hutch Bunny v1.7.0 rejects ICD-MAIN unconditionally: `execute_query` raises `NotImplementedError` for `code == "ICD-MAIN"` before any solver runs (upstream issue #30). A toggle could therefore never enable it, so shipping one would have been inert config advertising a capability that cannot exist. Note the enum value is the string `"ICD-MAIN"` (hyphen), not `ICD_MAIN`.

Current behaviour is unchanged and correct: an ICD-MAIN task hard-fails the run, the error lands in the artifact's `errors[]`, and nothing partial goes back to the Relay. Pinned by a regression test in 602ddd81.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
```diff
@@ -0,0 +1,86 @@
...
+    creds, schema = _resolve_credentials(options)
+    base_env = {k: os.environ[k] for k in (
+        "TASK_API_BASE_URL", "TASK_API_USERNAME", "TASK_API_PASSWORD", "TASK_API_TYPE",
+        "LOW_NUMBER_SUPPRESSION_THRESHOLD", "ROUNDING_TARGET",
+    ) if k in os.environ}
+    env = build_bunny_env(creds, schema=schema, collection_id=options.datasetId, base_env=base_env)
+
+    child = ChildResult.model_validate_json(_run_child(env))
+    if child.error:
+        raise RuntimeError(f"cohort_discovery hard-fail: {child.error}")
```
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1. Artifact writing is factored into a `_persist_artifact` helper used by both paths; on child failure the flow now persists an envelope carrying `errors=[{"code": "CHILD_FAILED", "detail": ...}]` **before** raising `RuntimeError`. `ArtifactEnvelope` gained an `errors` field. Covered by a test asserting the artifact is written *and* the error propagates.

plugins/flows/cohort_discovery/README.md
```diff
@@ -0,0 +1,78 @@
...
+| `TASK_API_USERNAME`                 | Relay Task API username.                                                                   |
+| `TASK_API_PASSWORD`                 | Relay Task API password.                                                                   |
+| `TASK_API_TYPE`                     | `a` = availability, `b` = distribution. (Bunny has no `c`.)                                |
+| `COLLECTION_ID`                     | The Relay collection id — **equal to the d2e `datasetId`** this poller serves.             |
+| `LOW_NUMBER_SUPPRESSION_THRESHOLD`  | Low-number suppression threshold applied to results (e.g. `10`).                           |
+| `ROUNDING_TARGET`                   | Rounding target applied to results (e.g. `10`).                                            |
+| `COHORT_DISCOVERY_ICD_MAIN_ENABLED` | Enable ICD_MAIN distribution tasks. **Default off.** When off, ICD_MAIN distribution tasks are excluded. |
+| `TASK_API_ENFORCE_HTTPS`            | Set **`false`** for a non-HTTPS Relay `TASK_API_BASE_URL`. Bunny's `DaemonSettings` enforces HTTPS by default. |
```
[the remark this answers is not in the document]
ohdsi-trex: Resolved by deleting the flag altogether (d3aabe91) — so both spellings are gone and the mismatch can't cause a deploy-time misconfiguration. See the `bunny_runner.py` thread for why: Bunny v1.7.0 cannot execute ICD-MAIN under any configuration. The README now documents the actual behaviour (hard-fail) under "Supported distribution task types", and the design spec/plan carry a dated correction.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
ohdsi-trex: Agreed and fixed in 859580e1. To confirm the detail: `duckdb` *is* present in `SupportedDatabaseDialects`, but it's not the dialect a cachedb dataset resolves to — that's `trex`, served over Trex's PostgreSQL wire protocol. The flow now resolves `trex` and no longer touches a DuckDB file. `duckdb` (along with `hana`/`bigquery`) is explicitly rejected with a message naming the supported set (postgres, trex).

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
ohdsi-trex: They're deploy-time env on the flow's runtime, not per-run params — the flow only reads `TASK_API_*`, `LOW_NUMBER_SUPPRESSION_THRESHOLD`, `ROUNDING_TARGET` from its environment and passes them to the Bunny child. The `DATASOURCE_DB_*` values are **not** deploy-time: they're resolved at run time from DBDao/TrexDao.

On the wiring, I traced it and think the premise needs one correction: `plugins/functions/package.json` configures the **edge functions**; it has no dataflow-worker env block. The flow runtime's env comes from the dataflow worker deployment — `docker-compose.yml` and `charts/d2e-services/templates/dataflow-worker-deployment.yaml`.

**I have not added the vars there yet** — it depends on the deployment model (the design assumes one poller deployment per dataset, since `COLLECTION_ID` = `datasetId`, which suggests per-deployment env rather than a worker-wide default). I've documented the location in the README and would like your call on where these should live. Happy to push it in this PR once you confirm.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_config.py
ohdsi-trex: Confirmed — it's `trex`. `TrexDao.dialect` returns `SupportedDatabaseDialects.TREX.value`, and `TrexDao._get_connection()` connects with psycopg2 to the Trex host/port using `dbname=cache_id`. `bunny_config.py` now keys on `_TREX_DIALECTS = {"trex"}`; the DuckDB set is removed.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_config.py
ohdsi-trex: Done in 859580e1, with one important nuance worth flagging.

TrexDao now resolves the cachedb connection details, and Bunny connects to the Trex **pgwire** endpoint with its own PostgreSQL client: `DATASOURCE_DB_DRIVERNAME=postgresql`, `DATASOURCE_DB_DATABASE=<cache_id>` (the pgwire dbname Trex routes on), host/port/user/password from `TrexDao.tenant_configs`. No DuckDB file path anywhere.

The nuance: Bunny cannot accept a TrexDao connection *object*. `hutch_bunny.core.db.base.BaseDBClient` requires a SQLAlchemy `Engine` + `inspector` and takes `Executable` statements, whereas TrexDao is psycopg2-based (raw SQL strings / `psycopg2.sql` composables). So TrexDao supplies the **connection details** and Bunny still owns SQL execution — which also keeps the design's "Bunny executes using its own database client" property. An adapter would have had to synthesise a SQLAlchemy engine anyway, so it would add nothing.

Two things I could not verify without a live Trex endpoint, flagged for deployment: (1) `SyncDBClient.__init__` runs `_check_tables_exist()`/`_check_indexes_exist()` through the SQLAlchemy inspector at construction, so postgres-dialect introspection has to work against Trex's catalogs; (2) TrexDao issues an explicit `USE <cache_id>` when `cache_id != database_code`, which a plain SQLAlchemy engine won't do — connecting with `dbname=cache_id` should cover it, but it needs a real run to confirm.

plugins/flows/cohort_discovery/cohort_discovery_plugin/types.py
ohdsi-trex: Done in 859580e1 — see my reply on the Copilot thread. Mutable defaults replaced with `Field(default_factory=...)`, plus a test that instances don't share default state.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_runner.py
ohdsi-trex: Not changed in this batch — flagging rather than silently deferring, because it interacts with a bug.

`max_iterations` is still hardcoded to 1 in `bunny_runner.py`. Two things to weigh before making it an input:

1. It contradicts the approved design spec, which fixes one poll per run ("not a persistent daemon"; cadence owned by the Prefect deployment schedule). Happy to change it, but it's a spec revision, not just plumbing.
2. `poll_for_tasks` counts **polls, not tasks**, and `PollingService` sleeps `POLLING_INTERVAL` (default 5s) at the end of *every* iteration — including the last — so `max_iterations=2` blocks ≥10s even when both tasks return instantly.
3. There's a latent bug that only triggers at >1: `_to_envelope` does `availability["count"] = r.count` per result, so a second availability task **silently overwrites** the first. The Relay would receive both, but the artifact would report only one. That must be fixed in the same change.

If you'd like it parameterised, I'd suggest adding it to `CohortDiscoveryOptions` (default 1, so current behaviour is preserved) and fixing the envelope to accumulate a list rather than overwrite. Say the word and I'll do both together.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
ohdsi-trex: Removed in 859580e1 — the multi-paragraph docstring is gone, replaced by a short one describing what `_resolve_credentials` returns. The `SimpleNamespace` DuckDB shim it was explaining is also gone.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
ohdsi-trex: Done in 859580e1 — the flow now persists the artifact (with the failure in `errors[]`) before raising, via a shared `_persist_artifact` helper. Details in my reply on the Copilot thread.

plugins/flows/cohort_discovery/README.md
ohdsi-trex: Yes — one "tick" = one flow run, i.e. every trigger of the scheduled deployment, and each run performs exactly one poll (`max_iterations=1`). The wording was ambiguous, so the README now says this explicitly instead of "tick".

plugins/flows/cohort_discovery/README.md
ohdsi-trex: Done in d3aabe91 — resolved by removing the flag rather than renaming it, so the README/spec mismatch is gone. The README now has a "Supported distribution task types" section documenting the real behaviour: ICD-MAIN cannot be enabled by any configuration at the pinned Bunny version, and such a task hard-fails with the error recorded in the artifact's `errors[]`. See the `bunny_runner.py` thread for the upstream detail.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_config.py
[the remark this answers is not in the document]
ohdsi-trex: Good catch — fixed in 859580e1.

The DuckDB branch is gone entirely: cachedb now goes through Trex's PostgreSQL wire protocol (per the discussion on this file), so there is no DuckDB datasource path left to mis-configure. `DATASOURCE_DB_SCHEMA` is now set in **every** supported branch, and `build_bunny_env` additionally hard-fails up front with a clear message if the schema is empty, plus a `_REQUIRED` completeness check over all six `DATASOURCE_*` values.

Confirmed your reasoning against the pin: `DaemonSettings.DATASOURCE_DB_SCHEMA` (`core/settings.py:83`) is declared with **no default**, so `Settings()` raises for every driver — not just DuckDB — when it is unset.

plugins/flows/cohort_discovery/cohort_discovery_plugin/tests/test_bunny_config.py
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1. The DuckDB mapping test is replaced by a Trex test that asserts `DATASOURCE_DB_SCHEMA` is set, along with the driver (`postgresql`), `DATASOURCE_DB_DATABASE` (the `cache_id` / pgwire dbname), and host/port/user/password.

The Postgres test now asserts `DATASOURCE_DB_SCHEMA` too, and there are new tests covering the regression you were guarding against: a missing schema hard-fails, an incomplete config hard-fails, and `duckdb` / `hana` / `bigquery` dialects are rejected.

plugins/flows/cohort_discovery/cohort_discovery_plugin/tests/test_integration.py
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1 — both `DATASOURCE_DB_SCHEMA` and `TASK_API_ENFORCE_HTTPS=false` are now set in the integration test's subprocess env, so it is runnable against the documented stub relay over plain HTTP.

One clarification added as a comment in the test: this path drives Bunny's own DuckDB client directly as a cheap local fixture. It is **not** the production path — production cachedb access is Trex pgwire via Bunny's PostgreSQL client.

plugins/flows/cohort_discovery/cohort_discovery_plugin/types.py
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1 — all mutable defaults now use `Field(default_factory=...)`, matching the convention in `dataflow_ui_plugin/types.py` that you pointed to. A test pins that two model instances do not share default state.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_runner.py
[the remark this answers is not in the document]
ohdsi-trex: You were right that the documented toggle and the code disagreed — but the resolution turned out to be the opposite of adding enforcement.

Checking the pinned Bunny (v1.7.0, `a4121dc`), `execute_query` **already rejects ICD-MAIN unconditionally**:

```python
if distribution_query.code == DistributionQueryType.ICD_MAIN:
    raise NotImplementedError("ICD-MAIN queries are not yet supported. ...")
```

(`core/execute_query.py:63`, before any solver runs; the enum value is the string `"ICD-MAIN"` with a hyphen.) So ICD-MAIN can never execute at this pin regardless of any flag, and a toggle would have been inert config advertising a capability that does not exist.

Rather than implement enforcement, the flag was removed everywhere (d3aabe91) and the behaviour documented: an ICD-MAIN task hard-fails the run, the error is captured in the persisted artifact, and nothing is returned to Relay. 602ddd81 adds a regression test pinning exactly that. Upstream tracking: https://github.com/Health-Informatics-UoN/hutch-bunny/issues/30

Note this diverges from the approved design spec §3.4, which assumed a working gate — the spec has been corrected in this PR. Shout if you'd rather keep a placeholder flag for when upstream lands support.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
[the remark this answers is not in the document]
ohdsi-trex: Fixed in 859580e1. The failure path now builds an envelope carrying a structured `errors[]` entry (`{"code": "CHILD_FAILED", "detail": ...}`), persists it via the shared `_persist_artifact` helper, and *then* raises `RuntimeError`. Success and failure paths share the same writer, so the artifact shape is identical either way.

`ArtifactEnvelope` gained an `errors` field for this. Test coverage asserts the artifact is written **and** the `RuntimeError` propagates, so the run still ends FAILED.

plugins/flows/cohort_discovery/README.md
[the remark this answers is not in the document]
ohdsi-trex: Resolved by deletion in d3aabe91 — rather than pick a name, the flag is gone from both sides.

As detailed in the `bunny_runner.py` thread: Bunny v1.7.0 rejects ICD-MAIN unconditionally, so neither `COHORT_DISCOVERY_ICD_MAIN_ENABLED` nor `COHORT_DISCOVERY_ENABLE_ICD_MAIN` could ever have done anything. Both spellings are removed from the README, the design spec, and the implementation plan, replaced by a short note describing the actual (hard-fail) behaviour. No deploy-time misconfiguration is possible now because there is no flag to set.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
ohdsi-trex: Confirmed and implemented in 859580e1 — thank you, this was the right call.

`SupportedDatabaseDialects` (`_shared_flow_utils/types.py:57`) does carry both `DUCKDB = "duckdb"` and `TREX = "trex"`, which is what led me to the DuckDB path originally. But for cachedb the DAO that actually resolves is `TrexDao` (`dialect` → `"trex"`), so the `duckdb` branch was unreachable in practice.

`_resolve_credentials` now returns the pgwire database name alongside the credentials, and `bunny_config` has a `trex` branch. `duckdb` is no longer an accepted dialect and raises `ValueError` naming the supported set (postgres, trex).

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
ohdsi-trex: Good question, and I think the premise needs correcting — leaving this thread open for your confirmation.

These are **deploy-time env on the dataflow worker runtime**, not per-run flow parameters: the flow reads `TASK_API_*`, `LOW_NUMBER_SUPPRESSION_THRESHOLD`, and `ROUNDING_TARGET` from `os.environ` and forwards them to the Bunny child. (The `DATASOURCE_*` values are different — those are resolved at run time from TrexDao/DBDao and never set by hand.)

On `plugins/functions/package.json`: I traced it and that file's `env` block scopes the **edge functions**, not the dataflow worker. The worker's runtime env comes from `docker-compose.yml` and `charts/d2e-services/templates/dataflow-worker-deployment.yaml`. So I don't think `functions/package.json` is the right place — but you know the deployment wiring better than I do, so please confirm.

I have deliberately **not** added `TASK_API_*` to compose/helm yet, because the values are per-dataset (one poller deployment per `COLLECTION_ID`) and I don't want to guess the model. Two things I need from you:
1. Confirm the correct location — compose/helm worker env, or somewhere else?
2. Given one deployment per dataset, should these be per-deployment env on the Prefect deployment rather than global worker env?

The README documents the required variables and flags this wiring as pending your confirmation.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_config.py
ohdsi-trex: Confirmed — it is `trex`.

Evidence: `_DAO_REGISTRY` in `_shared_flow_utils/dao/DBDao.py` maps `SupportedDatabaseDialects.TREX → TrexDao`, and `TrexDao.dialect` returns `"trex"`, so a cachedb dataset resolves to `TrexDao` and reports `trex` — never `duckdb`. `TrexDao._get_connection()` then connects over psycopg2 to the Trex host/port with `dbname=cache_id`.

`bunny_config.py` is updated accordingly: `_TREX_DIALECTS = {"trex"}` replaces the DuckDB set (859580e1).

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_config.py
ohdsi-trex: Done in 859580e1 — implemented as you asked, with one design note worth flagging.

TrexDao now resolves the connection details (host, port, user, password, and `cache_id` as the pgwire dbname), and Bunny connects to Trex over the **PostgreSQL wire protocol** using its own client (`DATASOURCE_DB_DRIVERNAME=postgresql`, `DATASOURCE_DB_DATABASE=<cache_id>`). No DuckDB file path anywhere.

The note: Bunny cannot be *handed* a TrexDao connection object. `hutch_bunny.core.db.base.BaseDBClient` requires a SQLAlchemy `Engine` plus an `inspector`, and its `execute`/`execute_and_fetch` take SQLAlchemy `Executable` objects, whereas TrexDao is psycopg2 with raw-SQL helpers. Since Trex speaks pgwire, pointing Bunny's existing `SyncDBClient` at it achieves what you asked without an adapter shim — TrexDao owns credential resolution, Bunny owns SQL execution.

Two things I could not verify without a live Trex endpoint, flagging so they are not a surprise at first run:
1. `SyncDBClient.__init__` runs `_check_tables_exist()` / `_check_indexes_exist()` through the SQLAlchemy inspector at construction — this assumes the postgres dialect can introspect Trex's catalogs.
2. TrexDao issues an explicit `USE <cache_id>` when `cache_id != database_code`; a plain SQLAlchemy engine will not, so catalog routing relies on connecting with `dbname=cache_id`.

Happy to adjust if you know either of those behaves differently in practice.

plugins/flows/cohort_discovery/cohort_discovery_plugin/types.py
ohdsi-trex: Done in 859580e1 — `distributions`, `raw`, and `results` now use `Field(default_factory=...)`, and `ArtifactEnvelope` gained an `errors` field (used by the new failure-path artifact). Replied in full on Copilot's thread.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_runner.py
ohdsi-trex: Deferred for now, and leaving this open — I'd like your call on the intended semantics, because it interacts with the approved spec and with a latent bug.

Why it wasn't changed in this round: design spec §2 lists "not a persistent daemon — each scheduled run performs exactly one poll cycle (`max_iterations=1`) and exits", with cadence owned by the Prefect deployment schedule. Making it a flow input contradicts that non-goal, so it felt like a spec change rather than a review fix.

Three things worth knowing before we decide, all verified against the pin:
1. **`max_iterations` counts polls, not tasks.** In `PollingService.poll_for_tasks`, a 204 (no task available) still consumes an iteration, so `max_iterations=2` yields 0, 1, or 2 tasks.
2. **There is an unconditional `time.sleep(POLLING_INTERVAL)` at the end of every iteration**, including the last (default 5s). So `max_iterations=2` blocks the child ≥10s even when both tasks return instantly — 5s of it dead wait after the final task.
3. **A real bug is currently masked by the hardcoded 1.** `_to_envelope` does `availability["count"] = r.count` per result, so two availability tasks in one run would silently overwrite the first count — the Relay would receive both results while the artifact reported only one. This must be fixed *together with* exposing the parameter, not after.

So: happy to plumb it through `CohortDiscoveryOptions` and the manifest's `parameter_openapi_schema`, but I'd want to fix the aggregation (list of results rather than last-wins) in the same change. Do you want:
- **(a)** the parameter exposed with default 1 + aggregation fixed, or
- **(b)** keep one poll per run and treat cadence purely as schedule config?

Let me know and I'll implement it.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
ohdsi-trex: Removed in 859580e1. The multi-paragraph docstring is gone, replaced with a short one describing what the function returns — including why the `database` value differs between `trex` (the `cache_id`, i.e. the pgwire dbname) and plain Postgres (`databaseName`). The `SimpleNamespace` DuckDB shim it described has been deleted along with the DuckDB path.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
ohdsi-trex: Done in 859580e1 — the artifact is now persisted with the structured error in `errors[]` *before* the `RuntimeError` is raised, so a failed run is inspectable in Prefect rather than only in the logs. Full detail on Copilot's thread.

plugins/flows/cohort_discovery/README.md
ohdsi-trex: Yes — one "tick" = one flow run = one trigger of the scheduled deployment. There is no in-process loop; each run performs exactly one poll (`poll_for_tasks(max_iterations=1)`) and exits, so cadence is entirely owned by the Prefect deployment schedule.

The wording was ambiguous, so the README now says this explicitly instead of relying on "tick" (859580e1).

plugins/flows/cohort_discovery/README.md
ohdsi-trex: Done in d3aabe91 — resolved by removing the flag rather than reconciling the two spellings, since Bunny v1.7.0 rejects ICD-MAIN unconditionally and the toggle could never have worked. Details on Copilot's thread.

plugins/flows/cohort_discovery/cohort_discovery_plugin/bunny_runner.py
ohdsi-trex: Closing the loop on my (a)/(b) question above: the team's decision is **(b)** — keep one poll per run for this iteration, so `max_iterations` stays fixed at 1 and is not exposed as a flow input.

That keeps the flow aligned with design spec §2 ("not a persistent daemon"; cadence owned by the Prefect deployment schedule). No code change was made here — current behaviour already matches the decision.

One consequence to record: the last-wins aggregation in `_to_envelope` stays as-is. It is unreachable while only one task can arrive per run, but if we ever revisit and parameterise `max_iterations`, that must be fixed in the same change or a second availability task will silently overwrite the first in the artifact.

Leaving this thread open for you to confirm or push back.

plugins/flows/cohort_discovery/cohort_discovery_plugin/flow.py
csafreen: You are right, the environment variables goes to compose/helm.
The cohort discovery plugin should have dataset id configurable per flow, not per deployment. It should run for any input datasetid.

csafreen:
Issues / suggestions

1. No subprocess timeout (flow.py:34, _run_child) — subprocess.run(...) has no timeout=. If the Bunny child hangs (network stall talking to Relay, a stuck DB query), the flow run blocks indefinitely with no hard-fail — which cuts against the "hard-fail everywhere" philosophy the rest of the code follows. Worth adding a timeout matched to the expected per-tick duration.
2. Duplicated branch logic in bunny_config.py:44-67 — the postgres and trex branches build an identical set of env vars (DATASOURCE_DB_DRIVERNAME/HOST/PORT/DATABASE/SCHEMA/USERNAME/PASSWORD); only the database argument's source and the trex-specific ValueError differ. Could collapse to one assignment block with an if dialect in _TREX_DIALECTS and not database: raise ... guard beforehand. Not urgent — the current form is more self-documenting via comments — but worth a look since duplication here means a future field addition has to be made twice.
3. Possible secret leakage into a persisted artifact — on child failure, flow.py:89-93 persists child.error (a string built in bunny_runner.py:45 from f"{type(exc).__name__}: {exc}") into the Prefect markdown artifact. If Bunny's or SQLAlchemy's exception formatting ever embeds the DSN (some drivers do, depending on version), DATASOURCE_DB_PASSWORD could end up in a durable, browsable artifact rather than just transient logs. Worth explicitly verifying what a real connection failure's exception string looks like at the pinned Bunny commit before relying on this in production — this is a step up in risk from just "logging" it, since artifacts persist and are more discoverable.
4. Minor: flow.py:1 and bunny_runner.py:2 combine multiple imports on one line (import os, json, subprocess) — inconsistent with typical PEP 8 style, though it does match the terse style already used elsewhere in this codebase, so it's a style nit rather than a real issue.

Zhimin-arya (approved):

csafreen (approved):

PR #2989, opened 2026-07-27
Add unit test suites for jobplugins, dataset and strategus-analysis plugin functions

### LINKED ISSUE #2521: Create unit test suites for initial three functions
Scope:
- jobplugins — input validation, core logic, error handling
- dataset — request parsing, input validation, core logic, error handling, auth case
- strategus analysis — input schema validation, analysis pipeline logic, result formatting, error propagation

ohdsi-trex:
## Summary

Adds `deno test` unit suites to three plugin functions that previously had none: `jobplugins`, `dataset`, and `strategus-analysis`. **42 tests, all passing.**

A shared harness under `plugins/functions/_shared/testing/` supplies Express request/response doubles, a `Trex` ambient-global shim, router lookup helpers, and an express-validator runner. Tests call route handlers directly with fake `req`/`res` and stub collaborators (API clients, TypeORM repositories) via `@std/testing/mock`. No live HTTP, no live database.

## Production change — one file pair, testability only

`dataset/index.ts` declared `export class DatasetRouter` *and* called `app.listen(8000)` at module scope, so importing the class started a server and hung the test process. The class is extracted verbatim into `dataset/router.ts`; `index.ts` is now bootstrap-only.

This is a **pure move**: the class body and import block are byte-identical to the original, and the registered route table was dumped before and after and matches exactly (8 layers, same order, `generateDatasetSchema` middleware intact). No handler logic, status code, or response string changed.

Two lazy-initialization changes considered during design were dropped as unnecessary — module-scope `Deno.env` reads are handled by importing a `_setup.ts` first (ES modules evaluate in import order), and TypeORM's `new DataSource()` does not connect until `.initialize()`.

## Coverage

| Plugin | Tests | Covers |
|---|---|---|
| `strategus-analysis` | 14 | study validation, analysis service logic, result enrichment, route error propagation |
| `jobplugins` | 21 | DQD request validators, controller validation, error handling, missing-auth path |
| `dataset` | 7 | request parsing, input validation, happy path, error handling, missing-token path |

Each test file was verified able to fail (assertion mutated, run, confirmed FAILED, restored).

## Known limitations

- **`--no-check` on `strategus-analysis` and `jobplugins`.** Both inherit pre-existing type errors that any importing test picks up: `datasource.ts` passes `PG__PORT` (a string) where TypeORM wants a number, and `PortalServerAPI.ts:280` references an out-of-scope `path`. Fixing them is a production change, kept out of scope. `dataset` type-checks cleanly.
- **Coverage is a proven pattern, not a sweep** — 1 of 17 `jobplugins` controllers and 2 of 8 `dataset` routes. Extending is mechanical repetition of the established shape.
- **These are characterization tests.** They pin current behaviour, including several questionable behaviours documented explicitly in test names/comments rather than silently fixed (see below).
- **No live edge-runtime verification.** Platform functions are pooled and do not hot-reload, so exercising the refactor against a served route needs an `alp-trex` restart.

## Pre-existing issues found while writing these tests (not fixed here)

1. `jobplugins/src/api/PortalServerAPI.ts:280` — `const path` is declared inside `try` but referenced in `catch`; any failure throws `ReferenceError` and destroys the original error.
2. `jobplugins/src/middlewares/DqdRequestValidatorMiddlewares.ts:25` — `.isUUID().notEmpty().withMessage(...)` attaches the message only to `notEmpty`, so a malformed UUID returns the unhelpful default `"Invalid value"`.
3. `dataset/router.ts` handles a missing `Authorization` header two different ways on the same router: `/cdm-schema/snapshot/metadata` constructs its API client outside the `try` (unhandled rejection, no response), while `/cohorts` constructs it inside (500). Neither returns 401.
4. `strategus-analysis` `validateStudyId` returns `true` when no token is supplied, so unauthenticated callers bypass study-existence validation.

Each is characterized by a named test asserting actual behaviour. Fixing any is a behaviour change deserving its own decision.

## Not included

The CI workflow that runs these suites (`.github/workflows/plugin-function-tests.yml`) is written and committed locally but **not in this PR** — the available token lacks the `workflow` OAuth scope. It needs to be pushed by someone with that scope, or added in a follow-up.


hengxian-jiang (approved):

suwarnoong (approved):

.github/workflows/plugin-function-tests.yml
```diff
@@ -0,0 +1,32 @@
...
+      matrix:
+        plugin: [strategus-analysis, jobplugins, dataset]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: denoland/setup-deno@v2
+        with:
+          deno-version: v2.x
+      - name: ${{ matrix.plugin }} tests
+        working-directory: plugins/functions/${{ matrix.plugin }}
+        run: deno task test
```
mwaiyee, line 32: Some tests are returning errors (eg [here](https://github.com/OHDSI/Data2Evidence/actions/runs/31569052736/job/94026939350?pr=2989#step:4:772)), are they expected?

hengxian-jiang: Yes, those are expected failure msg from the unit tests.

mwaiyee (approved):

p-hoffmann (approved):

PR #2998, opened 2026-07-28
Fix Pythia chat BAD_GATEWAY: set agent upstream base and default model in trex env

p-hoffmann:


p-hoffmann:
Superseded by #3012 — fork PRs get a read-only GITHUB_TOKEN, so every Docker Build job failed with 'denied: installation not allowed to Write organization package'. Reopened from an in-repo branch with the same commit.

PR #3016, opened 2026-07-29
enable tls in services

SantanM:
### Merge Checklist


suwarnoong (approved):

docker-compose.yml
```diff
@@ -184,8 +198,11 @@ services:
       # Property prefix updated for WebAPI p-hoffmann/openid branch: security.auth.openId.* (was security.oid.*)
       SECURITY_AUTH_OIDC_CLIENTID: ${LOGTO__D2E_APP__CLIENT_ID:-${LOGTO__ALP_APP__CLIENT_ID}}
       SECURITY_AUTH_OIDC_APISECRET: ${LOGTO__D2E_APP__CLIENT_SECRET:-${LOGTO__ALP_APP__CLIENT_SECRET}}
-      # Use Caddy's internal HTTP port (8080) for metadata discovery - avoids TLS issues
-      SECURITY_AUTH_OIDC_URL: "http://${PROJECT_NAME:-d2e}-caddy:8080/oidc/.well-known/openid-configuration"
+      # Discover OIDC metadata directly from Logto over HTTPS:3001 — same host/port as
+      # LOGTO__TOKEN_URL, so trex (token exchange) and WebAPI (discovery/JWKS) reach Logto
+      # consistently. The discovered issuer is still http://alp-caddy:8080/oidc (matches the
+      # token iss / LOGTO__ISSUER); only the fetch endpoint + jwks_uri are proper HTTPS on :3001.
+      SECURITY_AUTH_OIDC_URL: "https://${PROJECT_NAME:-d2e}-logto-1.${TLS__INTERNAL__DOMAIN:-alp.local}:3001/oidc/.well-known/openid-configuration"
```
p-hoffmann: The internal cert has no SAN for any of these hostnames.

This line (and the other 15 new `https://…${TLS__INTERNAL__DOMAIN:-alp.local}` URLs added in this PR) resolves to e.g. `d2e-logto-1.alp.local`. But the internal server cert's SAN is hardcoded, and it is `*.d2e.local`:

- `scripts/lib.sh:39` → `-addext "subjectAltName=DNS:*.d2e.local"`
- `scripts/lib.ts:240-241` → `*.d2e.local`, `d2e.local`

Neither file is touched by this PR. Note also that the two names are different variables: cert generation reads `TLS__INTERNAL__DOMAIN_NAME` (`lib.sh:12`, default `d2e.local`), while the hostnames here read `TLS__INTERNAL__DOMAIN` (default `alp.local`). Nothing writes `TLS__INTERNAL__DOMAIN` into `.env`, and the Helm chart hardcodes `value: alp.local` in three places (`charts/d2e-services/templates/d2e-deployment.yaml:915`, `charts/d2e-services/charts/d2e-core/templates/core-deployment.yaml:204` and `:347`).

I decoded the cert from a locally generated `.env` to confirm:

```
subject=CN=d2e.local
X509v3 Subject Alternative Name:
    DNS:*.d2e.local
```

Since this PR also installs the CA and removes the `rejectUnauthorized: false` escape hatches, verification is now strict — so trex→Logto, trex/worker→Prefect, portal→supabase-storage and Caddy's new `tls_trust_pool` reverse proxies should all fail with a hostname mismatch. The `.d2e.local` fallback in `services/alp-logto/post-init/src/middleware/logto.ts` looks like it was written against the other domain, which suggests the two halves were developed against different values.

Fix either way round: set `TLS__INTERNAL__DOMAIN` to `d2e.local` everywhere (compose default + Helm), or add `*.${TLS__INTERNAL__DOMAIN}` to the CSR SAN in both `lib.sh` and `lib.ts`.

Separately, the CI cert at `.github/workflows/docker-build-push.yaml:799-801` is generated with `-subj "/CN=d2e.cluster.local"` and **no** `-addext subjectAltName` at all, so it will not validate against anything once the chart starts verifying.

docker-compose.yml
```diff
@@ -1027,7 +1073,13 @@ services:
     entrypoint:
       - sh
       - -c
-      - npm run cli db seed -- --swe && npm run cli db alteration deploy latest && node packages/core/d2e-grants.mjs && npm start
+      - >-
+        mkdir -p /tmp/cert &&
+        printf '%s' "$TLS__INTERNAL__CRT" > /tmp/cert/cert.pem &&
```
p-hoffmann: Single `$` means Compose interpolates the private key into the container spec.

Every other cert writer in this file uses `$$` so the shell inside the container reads the value from its own environment: lines 115-117, and line 899. These two lines use a single `$`, so Compose substitutes the PEM at parse time instead.

Two consequences:

1. The private key ends up in the container's command — visible in `docker inspect` and `docker compose config`, and on the process command line.
2. Compose interpolation only reads the shell environment and `.env`; it does not read the container environment. So if these values ever arrive via `env_file` or Helm rather than `.env`, Compose substitutes empty strings and Logto boots with zero-byte `cert.pem`/`key.pem`, failing to start its HTTPS listener.

This service already has `<<: *x-tls`, so both variables are in its environment — changing `$` to `$$` on both lines is sufficient.

docker-compose.yml
```diff
@@ -958,6 +999,10 @@ services:
     platform: linux/amd64
     environment:
       ENV_TYPE: ${ENV_TYPE:-local}
+      # logto's admin API (LOGTO__ADMIN_SERVER__FQDN_URL) is now TLS-only with the
+      # shared self-signed internal cert; this container has no CA trust wiring,
+      # so skip cert validation for its outbound fetch() calls to logto.
+      NODE_TLS_REJECT_UNAUTHORIZED: "0"
```
p-hoffmann: This disables TLS verification process-wide in a container that handles secrets.

`alp-logto-post-init` posts `LOGTO_API_M2M_CLIENT_SECRET` (via `LOGTO__ALP_ADMIN_APP`), `LOGTO__CLIENTID_PASSWORD__BASIC_AUTH`, and every client secret in `LOGTO__CLIENT_APPS` to the admin API. Turning off verification makes those interceptable by anything that can win a DNS or routing race on the `alp` network — which is exactly the threat the rest of this PR is closing.

The comment says the container has no CA trust wiring, but that is easy to add: it currently has no `<<: *x-tls` in its environment. Adding the anchor and pointing `NODE_EXTRA_CA_CERTS` at a written-out `TLS__INTERNAL__CA_CRT` gives it real trust instead.

docker-compose-local.yml
```diff
@@ -105,7 +108,8 @@ services:
       EDGE_FUNCTIONS_MANAGEMENT_FOLDER: /app/edge-functions
       # Bind-mounted source plugins win over bundled ones with the same package name.
       # /usr/src/plugins-dev first so trex's own web/studio/notebook plugins load too.
-      PLUGINS_DEV_PATH: "/usr/src/plugins-dev:/usr/src/bundled-plugins:/usr/src/plugins"
+      # PLUGINS_DEV_PATH: "/usr/src/plugins-dev:/usr/src/bundled-plugins:/usr/src/plugins"
+      PLUGINS_DEV_PATH: "/usr/src/bundled-plugins"
```
p-hoffmann: This looks like a debug leftover that shouldn't ship.

Commenting out `PLUGINS_DEV_PATH` and replacing it with just `/usr/src/bundled-plugins` drops:

- `/usr/src/plugins-dev`, so bind-mounted edits under `plugins/` are silently ignored — which is the entire purpose of this override file, and
- `/usr/src/plugins`, i.e. trex's own web/studio/notebook plugins, per the comment immediately above.

It also seems unrelated to the TLS work in this PR.

plugins/functions/alp-usermgmt/src/api/_axios.ts
```diff
@@ -0,0 +1,206 @@
...
+      payload = "";
+    }
+  }
+
+  if (res.status >= 200 && res.status < 300) {
+    return {
+      data: payload,
+      status: res.status,
+      statusText: res.statusText,
+      headers: res.headers,
```
p-hoffmann: Returning the `Headers` object breaks every header read at the call sites.

Axios returned a plain object with lowercased keys, so call sites use bracket access. A `Headers` instance does not support that:

```js
const h = new Headers({'total-number': '250'})
h['total-number']         // undefined
Number(h['total-number']) // NaN
h.get('total-number')     // '250'
```

Concretely this breaks `LogtoAPI.fetchAllPages` (`src/api/LogtoAPI.ts:68`):

```js
const totalNumber = Number(result.headers['total-number'])
if (!totalNumber || allItems.length >= totalNumber) break
```

`Number(undefined)` is `NaN`, `!NaN` is `true`, so the loop always breaks after the first page and `fetchAllPages` silently returns at most `pageSize` (100) items. That affects `/api/users` via `getUsers()` (`LogtoAPI.ts:173`) — so the user list truncates at 100 users in any real tenant — as well as `/api/resources`, `/api/roles`, and the scope/role-scope lookups that `assignScopeToRole` uses to decide whether an assignment already exists.

Suggest converting to a lowercase-keyed plain object, e.g. `headers: Object.fromEntries(res.headers)`, in both the success path here and the error path on line 152. Applies to all 13 copies of this file.

plugins/functions/white-rabbit/src/api/_axios.ts
```diff
@@ -0,0 +1,206 @@
...
+      aborted ? `timeout of ${timeout}ms exceeded` : (e?.message || "Network Error"),
+      cfgForError,
+      aborted ? "ECONNABORTED" : "ERR_NETWORK",
+    );
+  }
+  if (timer) clearTimeout(timer);
+
+  let payload: any;
+  if (config.responseType === "arraybuffer") {
+    payload = new Uint8Array(await res.arrayBuffer());
```
p-hoffmann: `Uint8Array` where axios returned a `Buffer` corrupts the scan-report download.

`FilesManagerAPI.getRequestConfig()` passes `responseType: "arraybuffer"` and `getFile()` returns `result.data`, which `ScanDataService.scanReport()` hands straight to `res.status(200).send(result)` in `src/scan-data/scan-data.router.ts:62`.

Express's `res.send` only treats a payload as binary when `Buffer.isBuffer(chunk)` is true; anything else that is an object falls through to `res.json()`. I ran this against the exact version in `deno.lock` (`express@4.22.1`):

```
res.send(new Uint8Array([1,2,3,255]))
  → content-type: application/octet-stream; charset=utf-8
  → body: {"0":1,"1":2,"2":3,"3":255}

res.send(Buffer.from([1,2,3,255]))
  → content-type: application/octet-stream
  → body: 0x010203ff
```

So the downloaded scan report is JSON-encoded byte indices rather than the file. Suggest `payload = Buffer.from(await res.arrayBuffer())` here (or converting at the call site). Applies to all 13 copies of this file.

plugins/functions/alp-usermgmt/src/api/_axios.ts
```diff
@@ -0,0 +1,206 @@
...
+  } catch (e: any) {
+    if (timer) clearTimeout(timer);
+    const aborted = e?.name === "AbortError";
+    throw makeError(
+      aborted ? `timeout of ${timeout}ms exceeded` : (e?.message || "Network Error"),
+      cfgForError,
+      aborted ? "ECONNABORTED" : "ERR_NETWORK",
+    );
+  }
+  if (timer) clearTimeout(timer);
```
p-hoffmann: Nit (non-blocking): the abort timer is cleared as soon as `fetch` resolves, i.e. when response *headers* arrive, but before `res.text()` / `res.arrayBuffer()` on lines 123-127. A server that sends headers and then stalls the body will hang forever, whereas axios's `timeout` covered the whole response. Relevant given `axios.defaults.timeout = 30000` in the request-utils, and more likely to be hit now that these calls traverse TLS. Clearing the timer after the body is read (in a `finally`) would restore the old semantics.

plugins/functions/prefect/index.ts
```diff
@@ -9,26 +8,57 @@ if (!env.PREFECT_API_URL) {
...
+    if (!value) continue;
+    if (["host", "connection", "content-length"].includes(key.toLowerCase())) continue;
+    headers.set(key, Array.isArray(value) ? value.join(", ") : value);
+  }
+  headers.set("connection", "keep-alive");
+
+  let body: Buffer | undefined;
+  if (req.method !== "GET" && req.method !== "HEAD") {
+    const chunks: Uint8Array[] = [];
+    for await (const chunk of req) {
```
p-hoffmann: The request-body read sits outside the `try`, which only opens on line 40.

Express 4 does not await async handlers, so if the client disconnects mid-upload — or the request stream errors for any other reason — this `for await` rejects with nothing to catch it, producing an unhandled rejection rather than the 502 the `catch` below is there to return. Moving the body read inside the existing `try` is enough.

(I haven't verified whether the runtime installs an `unhandledrejection` handler, so I can't say for certain whether this takes the worker down or just logs — but either way it bypasses the error path deliberately written just below.)

services/alp-logto/post-init/src/middleware/logto.ts
```diff
@@ -1,5 +1,5 @@
 let LOGTO__ADMIN_SERVER__FQDN_URL =
-  process.env.LOGTO__ADMIN_SERVER__FQDN_URL || "http://alp-logto-1:3002";
+  process.env.LOGTO__ADMIN_SERVER__FQDN_URL || "https://alp-logto-1.d2e.local:3002";
```
p-hoffmann: Nit (non-blocking): this fallback doesn't match anything the deployment actually creates. With `PROJECT_NAME=d2e` the container is `d2e-logto-1`, not `alp-logto-1`, and the domain compose uses is `${TLS__INTERNAL__DOMAIN:-alp.local}`, not `d2e.local` — so both halves are off.

It only fires when `LOGTO__ADMIN_SERVER__FQDN_URL` is unset, and in that case it fails DNS resolution rather than degrading gracefully. Worth noting that the `.d2e.local` here is the correct SAN for the generated cert while compose uses `.alp.local` — see my comment on `docker-compose.yml`.

plugins/functions/analytics-svc/src/api/PsConfigServerAPI.ts
```diff
@@ -47,9 +46,6 @@ export default class PsConfigServerAPI {
                 "Content-Type": "application/x-www-form-urlencoded",
             }
         };
```
p-hoffmann, line 52: The `if (env.NODE_ENV === "development") { options.httpsAgent = new https.Agent({ rejectUnauthorized: false }) }` block was removed here with no replacement, and the new shim documents that `httpsAgent` is "accepted and ignored" — so there is no longer any way to reach a PS config server whose certificate isn't chained to the trusted internal CA.

That is the common case for this particular service, since it's typically external/HANA-side rather than one of our own containers. Any dev deployment relying on the old bypass will now fail the OAuth token POST with a TLS error where it previously succeeded.

Not blocking if that's an intentional call, but it should probably be replaced with something explicit (an opt-in env flag honoured by the shim, or trusting the relevant CA) rather than dropped silently.

p-hoffmann (changes requested):
Reviewed against the PR head. The direction is right; findings below, inline at the relevant lines.

The one I'd most encourage a look at is the first: as far as I can tell the internal cert's SAN doesn't cover any of the hostnames this PR now connects to over TLS, so the new `https://` internal hops may not verify. Evidence is in that comment.

- Internal cert SAN (`*.d2e.local`) vs. the hostnames used (`*.alp.local`) — `docker-compose.yml:205`.
- Logto's TLS private key gets interpolated by Compose into the container spec — `docker-compose.yml:1078`.
- `NODE_TLS_REJECT_UNAUTHORIZED: "0"` in `alp-logto-post-init`, which handles M2M and basic-auth secrets — `docker-compose.yml:1005`.
- Two behaviour changes in the fetch shim: response `headers` and `arraybuffer`.
- What looks like a debug leftover in `docker-compose-local.yml:112`.
- The Prefect proxy's body read sitting outside its `try`, and the removed dev-mode TLS bypass in `PsConfigServerAPI`.
- Two nits, marked as such inline: the shim's timeout scope and the post-init fallback host.

One general note: `_axios.ts` is committed as 13 byte-identical copies (all sha `e2e9b026`), so each of the shim fixes below has to be applied 13 times with nothing keeping them in sync. Worth considering `_shared/` instead. I also confirmed there are no remaining bare `axios` imports under `plugins/functions`, so the cutover itself is complete — but the now-dead `"axios": "npm:axios@1.6.0"` entries left in the various `deno.json` files invite someone to reintroduce the real client by accident.

p-hoffmann (approved):

mwaiyee (approved):

PR #3064, opened 2026-08-05
fix(portal): set cache_id to databaseCode for source datasets

### LINKED ISSUE #2877: Adding `source` type dataset should set databaseCode as cacheid
Image

When adding `source` type dataset, the cacheid should be set to `databaseCode` instead of `entity.id`
https://github.com/OHDSI/Data2Evidence/blob/67ef036d638a773d746b91196df7553ffd53473f/plugins/functions/portal/src/dataset/command/dataset-command.service.ts#L103

ohdsi-trex:
Closes #2877.

## Problem

Adding a dataset of type `source` produced a row whose `cache_id` names a DuckDB catalog that is never built.

The default branched only on **dialect**, never on **type**:

```ts
if (this.dialect === 'hana') this.cacheId = this.databaseCode
else if (this.id)           this.cacheId = sanitizeIdForCacheId(this.id)
```

Cache files are only built for WebAPI-managed datasets (`syncDatasetToWebApi` returns early for `type !== 'webapi'`), and a source dataset's cache lives on its *child* cache dataset — not on the source row. So a non-HANA source row pointed at a catalog nobody creates.

Because every consumer resolves `dataset.cacheId ?? dataset.databaseCode` (analytics-svc `dbsvc.ts`, `StudyDbCredential.ts`, d2e-webapi `trex.dao.ts`, terminology-svc `portal-api.ts`, parquet-export), a non-null-but-invalid `cacheId` **suppressed the databaseCode fallback** — queries hit a missing catalog instead of the source database, and `trex /attach` was asked to attach a file that does not exist.

Corroboration: the migration that introduced the column already backfills `cache_id = database_code`, so legacy and newly created rows disagreed.

## Change

- **`resolveCacheId()`** in `dataset.entity.ts` is now the single source of truth, used by all three write paths — the `@BeforeInsert` hook, the insert mirror (TypeORM `.insert()` skips hooks), and the `trex /attach` payload — so the persisted value and the attached value cannot drift.
- **The HANA rule is kept alongside the new one, not replaced.** `dialect === 'hana'` and `type === 'source'` hold for different reasons and neither subsumes the other: only the dialect branch covers `hana` + `webapi`, which is reachable because WebAPI management gates on the OMOP data model, not on dialect. A test pins this.
- **Snapshot rows** now derive their catalog from their own id, so a cache dataset owns its catalog instead of inheriting the source's.
- **Cache-file write target** (`jobplugins/CachedbController`) now resolves from the *cache* dataset. It previously read the source row, so once a source row holds its databaseCode the cache file would have been written to the source connection's own catalog while the cache dataset pointed at a catalog nobody wrote. Falls back to the source-derived value for legacy rows.
- **Backfill migration** repairing source rows created since `cache_id` landed. HANA rows are untouched.

## Behaviour matrix

| dialect | type | before | after |
|---|---|---|---|
| hana | `source` | databaseCode | databaseCode |
| hana | `webapi` | databaseCode | databaseCode |
| postgres | `source` | `sanitize(id)` ❌ | **databaseCode** ✅ |
| postgres | `webapi` | `sanitize(id)` | `sanitize(id)` |
| duckdb | `fhir` | `sanitize(id)` | `sanitize(id)` (unchanged) |

## Tests

New Deno suites, matching the `@std/testing` convention already used in portal:

```
plugins/functions/portal$     deno test --allow-all --no-check src/   →  7 passed (27 steps), 0 failed
plugins/functions/jobplugins$ deno test --allow-all --no-check src/   →  1 passed (7 steps),  0 failed
```

Each behaviour was mutation-tested rather than assumed: removing the `source` branch, removing the HANA branch, restoring the old snapshot inheritance, and reading the write target off the source row each fail the suite.

`deno check` reports no new errors; the pre-existing failures in `logger.ts`, `user-mgmt.api.ts` and `CachedbService` are unchanged from the base branch, and the migration typechecks clean.

## Not verified end-to-end

Portal is a bind-mounted platform function and there is no container runtime in the dev sandbox to restart `alp-trex`, so the running worker still serves the old code. Worth confirming on a stack that can restart: create a postgres `source` dataset with a cache dataset, then assert `cache_id = database_code` on the source row and the sanitized snapshot id on the cache row.

## Scope note

`type: 'fhir'` — the other `SourceDatasetType` — still gets `sanitize(id)`. The issue names only `source`, so that is left unchanged; a test pins current behaviour so the decision stays explicit.


trex/plans/issue-2877-source-dataset-cache-id.md
```diff
@@ -0,0 +1,167 @@
+# Issue #2877 — `source` type dataset gets an unusable `cache_id`
+
+- **Issue:** [OHDSI/Data2Evidence#2877](https://github.com/OHDSI/Data2Evidence/issues/2877) — "Adding `source` type dataset should set databaseCode as cacheid"
```
hengxian-jiang: Shall we remove the plan markdown file as part of the pull request

hengxian-jiang (approved):

p-hoffmann:
@copilot resolve the merge conflicts in this pull request

p-hoffmann (approved):

PR #3096, opened 2026-08-12
[Integrate-2410] apply authorization changes without waiting for the access token to expire

### LINKED ISSUE #2410: Permission -  User access not applied until relogin
Dataset: **Second datamart synpuf**

**Description**
When a user is added or deleted from the access list, the permission change does not take effect immediately. The user must log out and log back in before the updated access rights are applied.

**Expected Result:**
The permission should be effective right away without requiring the user to sign out and relogin.

**Actual Result:**
After adding or deleting a user from the access list, the permission change is not applied.

https://github.com/user-attachments/assets/b07a0c2f-8913-4af3-abce-1fa088bb80e4

ohdsi-trex:
Fixes #2410 — authorization changes did not take effect for a signed-in user until their access token expired (up to 1 hour) or they signed out.

## Problem

D2E carries authorization in the `roles` claim of a Logto access token. The claim is computed once, at mint time, and the token lives for an hour (`accessTokenTtl: 3600`). Administrative changes update the portal database and Logto, but not the token already in the user's browser.

Worse than a delay: on a request with `body.sync` set, `grant-roles-by-scopes` **reconciles the database to match the token**, so a stale token could write the *old* permission set back over a change an administrator had just made.

## Approach

Persist `usermgmt.user.authz_changed_at`. Every mutation that changes a user's authorization stamps it. `addUserObjToReq` — which already loads the user row on every request — compares the token's `iat` against that column and returns `401` with `X-D2E-Authz-Stale: 1` when the token predates the change.

Because `Repository.getOne` issues `SELECT *` and the middleware already fetches this row on every request, the new column arrives with the existing query — **no added round-trips on normal traffic**.

## What is in this PR

Tasks 1–6 of the plan:

- **Schema** — `authz_changed_at` migration, carried through `UserRepository`, the reducer, and the `User` entity
- **Comparison** — `authz-freshness.ts`, a pure `iat`-vs-`authz_changed_at` comparison with a configurable clock-skew allowance
- **Flags** — enforcement and skew settings in `env.ts`, surfaced in compose
- **Enforcement** — the freshness check in `addUserObjToReq`
- **Stale-reapplication guard** — `grant-roles-by-scopes` no longer reconciles roles from a token that predates the change (this is the data-corruption half of the bug)
- **Stamping** — `authz_changed_at` written at every mutation point, inside the surrounding transaction, covering all five administrative flows in the issue

Defaults are conservative: a `null` `authz_changed_at` means fresh, a misconfigured skew degrades to no allowance rather than locking anyone out, and enforcement sits behind a flag for staged rollout.

## Not in this PR

- **Task 7** — Logto session invalidation on revoke/delete
- **Task 8** — portal silent renewal and single retry, blocked on gates G1/G2 in the plan
- **Tasks 9–10** — end-to-end tests and staged rollout

Note the plan's checkboxes are not ticked; this PR delivers Tasks 1–6 as listed above.

## Plan and design

- Plan: `trex/plans/2026-08-12-authz-freshness-enforcement.md` (Revision 2 — scope narrowed to `add-user-object-to-req.ts`)
- Approved design: `trex/specs/2026-08-12-access-token-freshness-design.md` (Option A — silent renewal by default, forced re-login for revoke and delete)
- Analysis: `docs/issues/2410-token-claim-invalidation-replacement.md`

Both are committed on this branch so the intent sits next to the implementation.

## Test plan

- [x] `deno test --no-check` across the four new suites — **36 passed, 0 failed**, covering the freshness comparison, skew parsing, middleware enforcement, and transactional stamping
- [ ] Verify the migration applies against a live stack
- [ ] End-to-end coverage of the five administrative flows (Task 9)

Type-checking without `--no-check` reports pre-existing `Knex` namespace and `Buffer` errors in files this branch does not touch (e.g. `SetupService.ts`); they are unrelated to this change.

## Excluded — needs follow-up

An unrelated QA finding about **regression-test documentation** is **not** addressed here and needs separate follow-up. It is out of scope for this change and should be tracked on its own.

Separately, the devx QA browser review cannot meaningfully exercise this branch: the change is backend-only, the reviewer is handed the diff of the main workspace checkout rather than this branch, and the portal dev preview cannot authenticate because `env.js` is unavailable in dev so the OIDC config never loads. That also warrants its own follow-up.


SantanM:
The stale-token retry currently depends on `OidcLoginSilent` noticing a new `idToken` before it re-runs the user-group / role sync. But `refreshAuthToken()` only renews the access token, and that renewal may not rotate the ID token at all. If the ID token stays the same, the portal can skip the resync and remain in the stale auth state even after a successful refresh.

Could we trigger the resync directly from the stale-token retry path, or track token refresh in a way that doesn’t depend on `idToken` identity? That would make the recovery path deterministic.

hengxian-jiang (approved):

PR #3103, opened 2026-08-13
Add search window for non concept-set fields - closes #3102

### LINKED ISSUE #3102: Concept search overlay for concept-backed filter card attributes
**Problem**

In the cohort builder filter card, attributes that hold an OMOP concept identifier could only be filled through the small suggestion dropdown. There was no way to browse the vocabulary — the full-screen terminology overlay was reachable only from concept set attributes, via the + icon that creates a concept set.

**Change**

Text attributes configured with conceptIdentifierType now show a + ("Search Concepts") next to the dropdown. It opens the existing terminology overlay in CONCEPT_MULTI_SELECT mode, scoped by the attribute's domainFilter / standardConceptCodeFilter. Selected concepts are written back as ordinary tags.

Reopening the overlay pre-selects what is already on the constraint. Concept ids map directly; codes and names are resolved through the terminology service. Whatever the overlay shows is authoritative, so deselecting removes the tag.

**Configuration prerequisite**

The button and the domain preselection are driven entirely by the CDM config. An attribute needs conceptIdentifierType set (id, code or name), and domainFilter set if the overlay should open pre-filtered to a domain.

TimWalz:
Text attributes configured with conceptIdentifierType now show a + ("Search Concepts") next to the dropdown. It opens the existing terminology overlay in CONCEPT_MULTI_SELECT mode, scoped by the attribute's domainFilter / standardConceptCodeFilter. Selected concepts are written back as ordinary tags.

Reopening the overlay pre-selects what is already on the constraint. Concept ids map directly; codes and names are resolved through the terminology service. Whatever the overlay shows is authoritative, so deselecting removes the tag.

https://github.com/user-attachments/assets/3dee98de-995c-47ef-80e7-ba382854e76f

### Merge Checklist


tests/e2e/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "d2e-e2e-tests",
   "version": "1.0.0",
   "scripts": {
-    "test": "playwright test",
+    "test": "playwright test -u",
```
maggie-li-yd: With -u, every screenshot assertion silently regenerates inside the container and is then discarded, effectively making visual regression undetectable.

I'd suggest reverting this change.

plugins/ui/apps/vue-mri-ui-lib/src/lib/ui/app-tag-input.vue
maggie-li-yd, line 1: Latent bug discovered by claude:

### Latent overlay bug this PR makes reachable (fix belongs in `concept-sets`, not here)

**TL;DR** — `openConceptBrowser` is the first code path that awaits network I/O *before*
dispatching `alp-terminology-open`. That opens a window in which a second "+" can be
clicked, and the terminology listener mishandles a second dispatch: the overlay keeps the
**first** attribute's pre-selection but adopts the **second** attribute's `onClose`. Picks
made in that window are written into the wrong constraint.

**Mechanism** (`concept-sets/src/Terminology/`)

1. Idle: `props === null` → `TerminologyWithEventListener.tsx:33` returns `null`, so
   `<Terminology>` is unmounted.
2. First dispatch → `setProps(A)`, `setOpen(true)` → mounts →
   `useState(initialSelectedConcepts || [])` (`Terminology.tsx:392-394`) runs once.
3. Second dispatch while still open → `setProps(B)` → re-render, **no unmount**. The
   `useState` initializer doesn't re-run and nothing syncs it, so the selection stays A's —
   but `onClose` is now B's closure, carrying B's `identifier` and B's `resolvedValues`.

**Why this is latent on `develop`, not new code** — every existing dispatcher fires
synchronously inside its click handler (`app-tag-input.vue:101`,
`QueryFilterModern.vue:818` — `handleConceptSetAction` at `:608` isn't `async`,
`useTable1ConceptSetPicker.ts:146`, `ConceptSets.tsx:83`), and the overlay is
`<Drawer variant="temporary">` (`Terminology.tsx:102`), i.e. modal with a backdrop. Dispatch
and blocking happen in the same tick, so step 3 is unreachable today. This PR's
`await`-then-dispatch is what creates the gap.

**Recommended fix — two parts**

1. *Proper fix for `concept-sets`:* remount per open so the initializer
   re-runs and props can't get out of sync.

   ```tsx
   const [openSeq, setOpenSeq] = useState(0);
   // in listener: setProps(eventProps); setOpenSeq((n) => n + 1); setOpen(true);
   <Terminology key={openSeq} ... />
   ```

   Safe: the component already unmounts on every close, so remounting is the existing
   lifecycle. The only behaviour that changes is the second-dispatch-while-open case.

2. *Needed in this PR regardless,* since it can't ship on the current contract: a
   **module-scoped** in-flight guard in `app-tag-input.vue`. A per-component `data` flag
   does **not** close the window — two "+" buttons on one filter card are two component
   instances with two separate flags, which is exactly the cross-attribute case.

   ```js
   // module scope: one terminology overlay exists per page, so the guard must be shared
   // across instances.
   let conceptBrowserOpening = false
   ```

   Wire it to a disabled state on the button too — right now it looks dead for the whole
   resolution.

Not attributing (1) to this PR; flagging it because (2) is the part that has to land here.

maggie-li-yd (changes requested):

tests/e2e/package.json
TimWalz: Thanks, removed

Issue #3111, opened 2026-08-14
[D2E Redesign | DQD] DQD version is shown at the bottom right of the summary table


jessicasusilo-d4l:
### Description

The DQD summary table currently does not display which version of the Data Quality Dashboard was used to generate the results. The DQD version should be added to the bottom right of the summary table so users and admins can reference the exact version for reproducibility and debugging purposes.

---

### Acceptance Criteria

- The **DQD version number** is displayed at the bottom right of the DQD summary table
- The format: *"Data quality dashboard version: [version number]"* (e.g. *Data quality dashboard version: 2.0.0.100*)
- The version is read-only and reflects the version used for the most recent DQD run
- If no version information is available, the field is not shown

---

### Design Reference

- **Figma**: [Link to design file](https://www.figma.com/design/PEzbZHJJy6BKrwomMs8JYo/D2E-WebApp--MainFile-?node-id=1709-215281&t=eCB4oOacAzN0YgMP-11)

<img width="2880" height="1536" alt="Image" src="https://github.com/user-attachments/assets/83de0a3f-2dfa-44ce-a92d-8a6a99e2b10d" />

maksymtreshchov-d4l:
Checked on DEV, chrome latest, demo dataset

**QA Passed**

Attachments:

<img width="2557" height="1360" alt="Image" src="https://github.com/user-attachments/assets/22ca471c-c3af-444f-8b73-15a14656eae7" />

if no generation was run

<img width="2557" height="1360" alt="Image" src="https://github.com/user-attachments/assets/a930ec97-2ac6-4a86-b40a-a2b9f6c61084" />

Issue #3110, opened 2026-08-14
[D2E Redesign | DQD] DQD time stamp (generation timestamp) to be displayed on the left bottom of the summary table


jessicasusilo-d4l:
### Description

The DQD summary table currently does not display when the data quality dashboard was last generated. A generation timestamp should be added to the bottom left of the summary table so users can understand the recency of the data quality results.

---

### Acceptance Criteria

- A **generation timestamp** is displayed at the bottom left of the DQD summary table
- The timestamp format: *"Generated on [Month] [Day], [Year], [HH:MM] in [n] hours"* (e.g. *Generated on July 12th, 2016, 09:10 in 6 hours*)
- The timestamp is read-only and updates automatically when a new DQD run completes
- If no timestamp is available (e.g. DQD has never been run), the field is not shown

---

### Design Reference

**Figma**: [Link to design file](https://www.figma.com/design/PEzbZHJJy6BKrwomMs8JYo/D2E-WebApp--MainFile-?node-id=1709-215281&t=eCB4oOacAzN0YgMP-11)

<img width="2880" height="1536" alt="Image" src="https://github.com/user-attachments/assets/3c88f0d1-4703-4862-a3f4-9621e969a46c" />

maksymtreshchov-d4l:
Checked on DEV, chrome latest, demo dataset

**QA Passed**

- [x] A generation timestamp is displayed at the bottom left of the DQD summary table
- [x] The timestamp format: "Generated on [Month] [Day], [Year], [HH:MM] in [n] hours" (e.g. Generated on July 12th, 2016, 09:10 in 6 hours)
- [x] The timestamp is read-only and updates automatically when a new DQD run completes
- [x] If no timestamp is available (e.g. DQD has never been run), the field is not shown

Attachments:

<img width="2557" height="1360" alt="Image" src="https://github.com/user-attachments/assets/7f0b4b05-a82f-4ecb-b948-940efa123d5a" />

If no generation was run

<img width="2557" height="1360" alt="Image" src="https://github.com/user-attachments/assets/1530b598-28d3-45b3-a2ce-1d5dfeeb6f01" />

PR #3158, opened 2026-08-18
Expose DQD run metadata in the overview response

jerome-ng:
## Summary

- Extend the existing DQD overview endpoint with optional run metadata from the Prefect artifact.
- Return timing fields under a single `timing` object and unwrap the one-element arrays emitted by the R JSON serializer.
- Expose `dqdVersion` alongside the existing `total`, `validation`, and `verification` keys.
- Preserve the original three-key response when an older artifact does not contain timing or version metadata.

This provides the backend data needed for #3110 and #3111 without adding another UI API call.

## Endpoint

```text
GET /d2e/jobplugins/dqd/data-quality/flow-run/:flowRunId/overview?datasetId=:datasetId
```

Local example:

https://localhost:41100/d2e/jobplugins/dqd/data-quality/flow-run/b5db6c1d-e4dc-4b03-9a6c-b60c1c51539e/overview?datasetId=6784a0b0-38ad-4590-aeb6-180abed48322

## Response shape

The existing summary values are abbreviated below:

```json
{
  "total": { "...": "..." },
  "validation": { "...": "..." },
  "verification": { "...": "..." },
  "timing": {
    "startTimestamp": "2026-06-12 03:32:16",
    "endTimestamp": "2026-06-12 03:35:25",
    "executionTime": "3 mins",
    "executionTimeSeconds": 189
  },
  "dqdVersion": "2.6.0"
}
```

`timing` and `dqdVersion` are omitted when the artifact does not contain them. Individual fields inside `timing` are also optional.

## Root cause

The overview service already fetched and parsed the complete DQD artifact, but returned only the derived summary counts and discarded the artifact's timing and version metadata.

## UI integration note: timestamp timezone

The timezone of the stored DQD timestamp cannot be confirmed from the artifact itself. DQD creates the value using the R worker's system time and serializes it without a timezone offset or `Z` suffix.

The UI should therefore display the supplied timestamp value directly. It should not apply a browser/local conversion such as `+8`, append a timezone label, or otherwise reinterpret the value as a confirmed UTC timestamp.

## Validation

- `deno test --config plugins/functions/jobplugins/deno.json --no-lock --no-check --allow-env plugins/functions/jobplugins/src/services/__tests__/DqdService.test.ts`
- Focused test passed: enriched artifact metadata and legacy artifact fallback.
- Loaded locally by restarting `d2e-trex`; Trex returned a healthy internal status.

### DIFF
--- a/plugins/functions/jobplugins/src/services/DqdService.ts
+++ b/plugins/functions/jobplugins/src/services/DqdService.ts
@@ -50,11 +50,36 @@ export class DqdService {
       return null;
     }
 
-    const artifactData = JSON.parse(artifacts[0].data);
+    const artifactData = JSON.parse(artifacts[0].data) as IDataQualityResult;
     const checkResults = artifactData.CheckResults;
     const derivedResults = this.dataQualityOverviewParser.parse(checkResults);
+    const dqdVersion = artifactData.Metadata?.[0]?.dqdVersion;
+    const timing = {
+      ...this.getArtifactTimingValue(
+        "startTimestamp",
+        artifactData.startTimestamp
+      ),
+      ...this.getArtifactTimingValue("endTimestamp", artifactData.endTimestamp),
+      ...this.getArtifactTimingValue(
+        "executionTime",
+        artifactData.executionTime
+      ),
+      ...this.getArtifactTimingValue(
+        "executionTimeSeconds",
+        artifactData.executionTimeSeconds
+      ),
+    };
 
-    return derivedResults;
+    return {
+      ...derivedResults,
+      ...(Object.keys(timing).length > 0 ? { timing } : {}),
+      ...(dqdVersion ? { dqdVersion } : {}),
+    };
+  }
+
+  private getArtifactTimingValue<T>(key: string, value?: T | T[]) {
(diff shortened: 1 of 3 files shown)


jerome-ng:
UI integration note: the timezone of the DQD timestamp cannot be confirmed from the artifact. The value is based on the system timezone used by the R worker and is serialized without an offset. Please display the supplied value directly in the UI; do not apply a conversion such as +8, append a timezone label, or otherwise reinterpret it as confirmed UTC.

jerome-ng:
Additional API contract note for the UI integration:

- `dqdVersion` is added only when `Metadata[0].dqdVersion` is present and non-empty.
- `timing` is added only when at least one timing value exists.
- Within `timing`, `startTimestamp`, `endTimestamp`, `executionTime`, and `executionTimeSeconds` are each added independently when available. The one-element arrays emitted by the R serializer are unwrapped to scalar values.
- If the artifact contains none of this metadata, the response remains the existing `{ total, validation, verification }` shape. It will not include an empty `timing: {}` object or an empty `dqdVersion` value.

The UI should therefore treat both top-level additions and every nested timing field as optional.

Issue #3124, opened 2026-08-14
[D2E Redesign | Exploration | Action] Delete - open a delete confirmation dialog


jessicasusilo-d4l:
### Description

The Delete action allows users to permanently remove a Data Exploration from the list. Logic of the delete pop up will be similar to the current pop up, but need to change the style to the new dialog component

---

#### Entry Point

- **Delete** is accessible from the **More (⋮)** dropdown menu on the exploration card
- The Delete label is styled in red to indicate a destructive action

#### Delete Confirmation Dialog

- Clicking Delete opens a **"Delete filter?"** confirmation dialog
- The dialog contains:
  - Dialog title: **"Delete filter?"**
  - Warning message: *"Deleting this saved filter will delete any access point that you generated for it. This action cannot be undone."*
  - **Cancel** button — closes the dialog without deleting
  - **Yes, delete** button (destructive, red) — permanently deletes the exploration
  - A close (×) button in the top right
- Once confirmed, the exploration card is removed from the list immediately

---

### Design Reference

**Figma**: [Link to design file](https://www.figma.com/design/PEzbZHJJy6BKrwomMs8JYo/D2E-WebApp--MainFile-?node-id=1810-243152&t=eCB4oOacAzN0YgMP-11)

<img width="4346" height="6510" alt="Image" src="https://github.com/user-attachments/assets/cbd601b4-6556-4e8d-a47e-75f609c80277" />

Issue #3122, opened 2026-08-14
[D2E Redesign | Exploration | Action]  Rename — opens a dialog to rename the exploration


jessicasusilo-d4l:
### Description

The Rename action allows users to update the name of an existing Data Exploration from the More action menu on the exploration card. Behavior should follow the current behaviour, but dialog to follow the new dialog component

---

#### Entry Point

- **Rename** is accessible from the **More (⋮)** dropdown menu on the exploration card

#### Rename Dialog

- Clicking Rename opens a 

**"Rename exploration name"** dialog
- The dialog contains:
  - Dialog title: **"Rename exploration name"**
  - A required text input field labelled **"Exploration name *"** pre-populated with the current exploration name
  - **Cancel** button — closes the dialog without saving
  - **Rename** button (primary) — saves the new name and closes the dialog
  - A close (×) button in the top right
- The Rename button is disabled if the field is empty
- Once confirmed, the exploration card updates immediately to reflect the new name

---

### Design Reference
**Figma**: [Link to design file](https://www.figma.com/design/PEzbZHJJy6BKrwomMs8JYo/D2E-WebApp--MainFile-?node-id=1810-236905&t=eCB4oOacAzN0YgMP-11)

<img width="1469" height="696" alt="Image" src="https://github.com/user-attachments/assets/48c34dc7-98e8-4968-9d78-d9fcb1e49bc9" />

Issue #3118, opened 2026-08-14
[D2E Redesign | Exploration | Action] Materialize


jessicasusilo-d4l:
### Description

The Materialize action on the Data Exploration card allows users to trigger materialization of their cohort against the data source. Upon clicking, a dialog is shown prompting the user to optionally add a description before proceeding.

---

#### Entry Point

- The **Materialize** action is accessible from:
  - The **Materialize** button shown on the card when status is **Not run yet**
  - The first icon in the **quick action bar** at the bottom of the card (with a tooltip label *"Materialize"* on hover)

---

#### Materialise Cohort Dialog

- Clicking the Materialize action opens a **Materialize cohort** dialog
- The dialog contains:
  - Dialog title: **"Materialize cohort"**
  - A text input field labelled **"Cohort materialization description"** — this field is **optional**
  - **Cancel** button — closes the dialog without triggering materialisation
  - **Materialize** button (primary) — confirms and triggers the materialization process
- A close (×) button is available in the top right of the dialog (same behaviour as "Cancel")

---

### Design Reference

**Figma**: [Link to design file](https://www.figma.com/design/PEzbZHJJy6BKrwomMs8JYo/D2E-WebApp--MainFile-?node-id=1798-193374&t=eCB4oOacAzN0YgMP-11)

<img width="4306" height="3004" alt="Image" src="https://github.com/user-attachments/assets/65a3f00a-ab34-4c37-9010-e31c9d551bdf" />

Issue #3116, opened 2026-08-14
[D2E Redesign] Data exploration


jessicasusilo-d4l:
### Description

Currently, D2E cohorts and Atlas cohorts are displayed together in the same cohort list. As part of the D2E redesign, D2E cohorts are separated into a dedicated **Data Exploration** section, accessible via the **Exploration** page in the top navigation. The concept of "Data Exploration" replaces what was previously called "D2E cohort"; users can quickly explore their data by creating and materialising explorations using the D2E cohort builder without the overhead of managing a full cohort definition.

---

#### Page Header

- A **D2E · DATA EXPLORATION** breadcrumb label is shown above the page title
- Page title: **"Data Exploration"**
- Page description: *"Get quick insights about your data source, an easy way to validate whether a cohort is possible before defining it further for your study. Save your exploration filter and reuse it across the site, then materialise it against different data sources."*
- A **Data source** selector dropdown is available in the top right

#### Page Toolbar

- A **Search** bar is available to search across explorations. 
  - User should be able to search based on the name of the exploration (PA)
- A **Filters** button is available. Filter will be covered under **[this ticket](https://github.com/OHDSI/Data2Evidence/issues/3150)**
- A **Sort by** dropdown — default sort: **Last updated**
  - User should be able to sort card by Last updated, Name A-Z, Name Z-A
- A **+ Start new exploration** primary action button is available in the top right > this will bring user to open the D2E cohort builder page to start their exploration

#### Exploration Card

Each exploration is displayed as a card containing the following:

##### **Top section:**
- Exploration name (bold)
  - If the name is longer than the card width, it is truncated with an ellipsis (e.g. *"SNRI Users Researcher trial Hel..."*) and a tooltip shows the full name on hover
- Person count + **persons** label (large, bold) — shown when user has materilised particular exploration in data source selected
- Status badge (top right) - **This status badge can be hidden for now until we implement** **[this ticket ](https://github.com/OHDSI/Data2Evidence/issues/3117)**
- A **checkbox** in the top right corner for bulk selection - **Bulk action flow will be covered under** **[this ticket](https://github.com/OHDSI/Data2Evidence/issues/3127)**

##### **Materialization details:**
- Last Materialized on — date (shown in navy blue)
- Exploration ID — numeric ID (shown in navy blue)
- Description — short description text (shown in navy blue)

##### **Exploration bookmark section:**
- Created by — username
- Last updated — date
- Version — version number

[...]

PR #3205, opened 2026-08-26
feat(vue-mri): reskin the rename, delete and materialize dialogs

khairul-syazwan:
Stack 2 of 3 for the D2E data-exploration redesign (OHDSI/Data2Evidence#3116).

Reskins the three existing dialogs onto the d2e component library:
- Rename dialog (#3122)
- Delete dialog (#3124)
- Materialize dialog (#3118)
- Rebuild `UnsavedChangesDialog` on `D2eDialog`.

Notes:
- Dialogs use the Modal S (540) width, matching the Figma frames. The materialize dialog stays at S; a card sits behind it, not inside it.
- The scrim no longer dismisses. Escape still closes.
- Draft: the rename empty-name behavior (#3122) needs a final decision from design. The delete dialog uses the `danger` variant, which is pending a design decision on the destructive button category.


khairul-syazwan:
Superseded by #3259, which carries the same branch tip onto the tokens base. Closing this earlier draft so the stack has one modals PR.

PR #3251, opened 2026-09-02
feat(ui): build @d2e/ui as a distributable component library

khairul-syazwan:
Stack 2 of 5 for the D2E data-exploration redesign (OHDSI/Data2Evidence#3116). Stacked on #3204 — review that first.

Packages `@d2e/ui` as a built library so the portal can share it once it moves to Vue, instead of every consumer compiling the Vue SFCs itself.

**What changed**
- vite lib mode -> `dist/index.js` + `dist/index.css`, types via `vue-tsc`. Not the rollup/babel chain `libs/portal-components` uses: that is rollup 2 plus six babel packages for React/TSX, and vite handles Vue SFCs natively.
- `vue` and `vuetify` stay external as peer dependencies.
- Components import the Vuetify pieces they use, so a consumer no longer needs `vite-plugin-vuetify`.
- Entry points move to `dist`. `./tokens.css` stays on source — it is generated from `tokens.ts` and guarded by `tokens:check`.
- `scripts/verify-dist.mjs` checks the artifact, wired into the CI step added in #3204.

**Two things worth a reviewer's attention**

Adding the Vuetify imports broke three unit tests: they imported their constant maps from the `.vue` files, so they began loading Vuetify's CSS under vitest's node environment. The maps moved into plain modules (`buttonVariants`, `iconButtonSizes`, `statusChipVariants`, `explorationCardStatus`), following the existing `dialogSizes.ts` pattern. Public API unchanged.

The obvious smoke test — importing `dist` in plain node — cannot work, because vuetify's ESM pulls in `.css`. `verify:dist` checks statically instead and asserts *positively* that vue and vuetify are still imported: when a peer gets bundled it stops appearing as an import at all, so checking only for unexpected imports misses the case that matters. Verified by dropping vuetify from `external` — it fails on both the missing import and a 197 kB bundle against the 150 kB ceiling.

**Not in scope**
The vite aliases in `apps/vue-mri-ui-lib` stay. The package is private and unpublished, so the ATLAS job's `npm install --workspaces=false` cannot resolve it from the registry regardless of a dist. The application keeps consuming source, which also preserves hot reload.

**Validation**
25 library tests, `tokens:check`, `verify:dist`, all three vue-mri build variants, explorer 8 stories / 30 variants, and every component re-checked in the browser with no console errors.

### DIFF
--- a/.github/workflows/ui-test-vue.yml
+++ b/.github/workflows/ui-test-vue.yml
@@ -86,3 +86,6 @@ jobs:
         # fails if tokens.css has drifted from tokens.ts
         bun run tokens:check
         bun run test
+        # build the distributable and check its contract
+        bun run build
+        bun run verify:dist
--- a/plugins/ui/bun.lock
+++ b/plugins/ui/bun.lock
@@ -550,6 +550,7 @@
         "vite": "^6.4.2",
         "vitest": "^4.0.18",
         "vue": "^3.5.17",
+        "vue-tsc": "^2.2.12",
         "vuetify": "3.12.0",
       },
       "peerDependencies": {
--- a/plugins/ui/libs/d2e-ui/.gitignore
+++ b/plugins/ui/libs/d2e-ui/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
 .histoire/
+dist/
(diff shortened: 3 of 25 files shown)


khairul-syazwan:
Folded into #3204 rather than reviewed separately.

The packaging change was small — 25 files, +323/-121 — against #3204's +7448, and 8 of those files are components #3204 itself introduces. Splitting them meant a reviewer read those SFCs twice: once with the constant maps inline, then again with the maps extracted for the packaging refactor. The final state is clearer reviewed once.

The commits are unchanged and now sit on `khairul-syazwan/d2e-ui-library`. Nothing is lost; #3204 was fast-forwarded, not force-pushed. #3253 is retargeted onto #3204 and the stack renumbered to 4.

PR #3256, opened 2026-09-02
fix(pa): reset chart axes from config defaults

### LINKED ISSUE #3075: Cohorts - Barchart displays incorrectly  after reseting filter
Dataset HANA_CDMSYNPUF1K

**Step:**
1. Create new D2E  → Age attribute is set as the default X1 axis
2. Switch X1 to Gender
3. Click on Reset filter button → X1 axis return to default value : Age
4. Observe the bar chart

**Expected Result:**
The bar chart should displays properly with Age as default X1 axis

**Actual Result:**
It displays with with Gender as the X1 axis instead of resetting to Age.
- Swtich X1 To Gender then reset

https://github.com/user-attachments/assets/31e63075-beb9-454b-9a50-f07236c17cda

- Swtich X1 To another attribute then reset

https://github.com/user-attachments/assets/a59f4e63-d47e-4940-ad56-284d627a6b1a

jerome-ng:
## Summary

- restore complete chart-axis selections from the active PA configuration, including default bin sizes
- clear stale attribute and bin-size state from unused default axis slots
- make chart reset a request-batched transaction and fire one refresh after the final state is committed
- await the reset transaction from the filters footer

## Root cause

The reset path resolved the configured initial axes, but only merged `key`, `filterCardId`, and `attributeId` into existing axis models. Selection-dependent properties such as the previous attribute's bin size could survive. The footer also reset the query and chart independently without awaiting a single final refresh, allowing the selector state and rendered chart request to diverge.

## Verification

- `bun run test:unit -- src/store/modules/__tests__/chart.reset.test.ts` — 4 passed
- `bun run test:unit` — 941 passed, 3 skipped
- `bun run build` — passed
- hot-deployed vue-mri to both local Trex resource roots
- Playwright regression: changed X1 away from the local config default, reset, and verified one final barchart request whose axis IDs/bin sizes matched the initial configured state; the rendered Plotly traces also matched the initial chart

The local stack exposes only `Demo dataset`, whose defaults are Gender/Age, rather than the issue's `HANA_CDMSYNPUF1K` Age-first configuration. The runtime check therefore exercised the same config-driven reset with the locally available defaults.

Closes #3075


plugins/ui/apps/vue-mri-ui-lib/src/components/FiltersFooter.vue
```diff
@@ -283,7 +283,7 @@ export default {
     },
   },
   methods: {
-    ...mapActions(['fireBookmarkQuery', 'loadbookmarkToState', 'resetChart', 'queryReset']),
```
jerome-ng, line 288: queryReset is done in resetChart instead

plugins/ui/apps/vue-mri-ui-lib/src/store/modules/chart.ts
```diff
@@ -318,33 +318,36 @@ const actions = {
...
+    const initialAxis = mriFrontendConfig.getInitialAxisSelection()
+    for (let i = 0; i < Constants.MRIChartDimensions.Count; i += 1) {
+      let filterCardId = ''
+      let key = ''
+      if (initialAxis && initialAxis[i] && initialAxis[i] !== 'hc.mri.pa.ui.lib.Selection.Invalid') {
+        const axisValue = initialAxis[i].split('.')
+        key = axisValue.pop()
+        axisValue.pop()
+        filterCardId = axisValue.join('.')
+        const defaultBinSize = mriFrontendConfig.getAttributeByPath(initialAxis[i])?.getDefaultBinSize?.()
```
jerome-ng, line 332: default bin size now taken from the config

khairul-syazwan (approved):

Issue #3291, opened 2026-09-08
[D2E Redesign | Exploration] Pagination and empty state


khairul-syazwan:
### Description

The Data Exploration page's grid has no pagination, and its empty state is a single bare sentence with no illustration. This ticket covers two frames from the same Figma board that the original planning did not cover:

- A pagination bar under the exploration grid: a **Row per page** control, a **1-12 of 43** style count, and first / previous / next / last page controls. The grid is client-side and pages client-side; the list endpoint takes no `page` or `limit` parameter.
- A designed empty state: a 140x140 illustration above a title and a body line, replacing today's one-line text. The same empty state currently renders for three different situations (no explorations at all, a search that matched nothing, a filter that matched nothing); only the first is "No explorations yet" — the other two need their own copy so they do not read as data loss.

---

#### Pagination

- A pagination bar sits under the exploration grid, showing a page-size control, the current range and total count, and first/previous/next/last controls.
- The grid shows one page at a time; the default page size is 12.
- First and previous are disabled on the first page; next and last are disabled on the last page.
- Changing the page size, the search, a filter, or the sort returns to page 1.
- The bar does not render when nothing matches.

#### Empty state

- The empty state shows an illustration above a title and a body line.
- An account with no explorations, a search that matches nothing, and a filter that matches nothing each show their own copy.
- The page toolbar stays visible and usable while the empty state is shown.

---

### Design Reference

**Figma**: file `PEzbZHJJy6BKrwomMs8JYo`, "Pagination" board `3684:161320` and "Empty state" board `3684:161324`.

PR #3292, opened 2026-09-08
feat(vue-mri): pagination and the designed empty state for Data Exploration

khairul-syazwan:
## Summary

Adds the two frames from the exploration page's design that earlier planning
never covered:

- A pagination bar under the grid, client-side, default page size 12, with
  first/previous/next/last controls and a `Row per page:` size selector.
- The designed empty state (140x140 illustration, title, body), replacing the
  bare "No explorations yet." sentence, split into three copies so a search or
  a filter that matches nothing no longer reads as "you own nothing".

Ticket: #3291 (filed for this work, no ticket existed before).

## Details

Blueprints: `docs/projects/vue-mri-ui/pr12/01-pagination.md` and
`docs/projects/vue-mri-ui/pr12/02-empty-state.md`. Deviations recorded in
`docs/projects/vue-mri-ui/DECISIONS.md` section 13:

- The page-size box is a native `<select>`, not `D2eSelect` — the frame's box
  has neither `D2eSelect`'s height nor its floating label.
- `PAGE_SIZES` is `[12, 24, 48]`, not the frame's shown `12` alone and not a
  conventional preset like `[10, 25, 50, 100]` either: the grid is four
  columns wide, and 12/24/48 are the multiples of 4, so every full page fills
  complete rows.
- The empty state's two "no match" situations (search, filter) get separate
  i18n keys per the blueprint, even though today's English copy is identical
  for both.
- The empty state reuses the page's existing status-container class rather
  than the frame's own block styling (different padding, a bottom border);
  neither is an acceptance criterion here.

## Fixed along the way

**A code review** found two defects before this was first opened: the
pagination bar could strand on a stale page if the list shrank through a path
the reset-on-change watcher doesn't cover (e.g. deleting the last card on a
page), and the page-size select had no programmatic label for a screen
reader.

**Manual verification** (seeding 13 dummy explorations through the app's own
save flow to get past one page, then clicking through both features at a real
viewport) found four more:

- The empty state was not centred — a `class` passed to `ExplorationEmptyState`
  fell through onto its own root and merged with its fixed-width layout class
  on the same element, so centering only applied to the illustration/text
  within that box, not the box within the page.
- The page was not clamped to the viewport — the pagination bar scrolled out
  of view with the rest of the page instead of staying fixed under the grid.
  Header, toolbar and the bar are now pinned; only the grid/status region
  scrolls, with `justify-content: safe center` so overflowing content in a
  short viewport falls back to `flex-start` instead of clipping unreachably,
  and `flex-shrink: 0` on the fixed chrome so it's the scrollable region that
  gives way first.
- `pageLabel` reinvented comma-thousands formatting that already exists as
  `NumberUtils.formatNumber`, used elsewhere in the app; swapped to reuse it.

## Test plan

- [x] `npx vitest run src/components/helpers/__tests__/explorationPaging.test.ts` — 13 tests
- [x] `npx vitest run` — 1009 passing
- [x] `npx vitest run --coverage` on Node 20 (the actual CI command) — 1009 passing
- [x] `npx vite build`, `--config vite.config.atlas.ts`, `--config vite.config.atlas-app.ts` — all green
- [x] Hot-deployed to trex and verified live with Playwright and by hand: baseline pagination
      (`1-3 of 3`, all four controls correctly disabled with 3 demo records); a search matching
      nothing shows the search empty-state copy, centred, and hides the bar; filtering to
      `Materialized` shows the filter copy (taking precedence over an empty search) and hides
      the bar; clearing the filter restores the grid and the bar; with 16 records (13 seeded
      dummies), page 2 reads `13-16 of 16` with Next/Last correctly disabled; scrolling the grid
      to its end leaves the pagination bar's screen position unchanged and the page itself never
      scrolls.
- [ ] Lint not run — `plugins/ui` hoists eslint 7.32, which cannot read the app's flat config
      (pre-existing, tracked stack-wide).

`no-mistakes` was skipped per the skill's own worktree note: it does not
follow a treehouse worktree and targets `develop`, not this stack's parent.


khairul-syazwan:
Fixes the untranslatable pagination count (ROADMAP 4.10). `pageLabel()` was composing "1-12 of 43" in the pure helper with a hard-coded `"of"` and the literal `"0 of 0"`, which no i18n key could translate. Split it into a locale-neutral `pageRange()` that returns the comma-grouped start/end/total, and compose the final label in `ExplorationPagination.vue` from two new keys — `MRI_PA_EXPLORATIONS_PAGINATION_RESULT` and `MRI_PA_EXPLORATIONS_PAGINATION_EMPTY` — so a translator owns the word order and punctuation. Keys added for en, de and zh.

Verified: `explorationPaging.test` 13/13 pass; prettier clean on the touched helper, component and test; the vue-mri app builds.

jerome-ng (approved):

PR #3295, opened 2026-09-08
Plugin to calculate eq-5d-5l index value

### LINKED ISSUE #3048: Create a prefect job to calculate index value with country specific weights for Eq-5d-5l responses
Calculate and store EQ-5D-5L index values in the Measurement table.

Implement a job that calculates EQ-5D-5L index values (utility scores) from raw EQ-5D-5L responses using the appropriate country-specific value set, and writes the results to the Measurement table.

Requirements:

Input: EQ-5D-5L responses (5 dimensions × 5 levels: mobility, self-care, usual activities, pain/discomfort, anxiety/depression), plus a country code associated with the dataset.
Single country per dataset: All responses within a given dataset/run use one country code — the job does not need to support mixed-country batches in a single execution.
Value set application: Use the correct EuroQol country-specific value set to convert each 5-digit health state (e.g., 11223) into a utility index score. Value sets to be sourced from the [EuroQol value sets page](https://euroqol.org/information-and-support/resources/value-sets/).
Output: Write the calculated index value to the Measurement table, linked to the relevant subject/encounter/response record.
Re-run / overwrite behavior: If the job is re-run for the same dataset using a different country code, the newly calculated index values must overwrite the previously stored values in the Measurement table (not create duplicate rows).

csafreen:
EQ-5D-5L Index Calculation Plugin — Technical Deep Dive
Status: Temporary/working document. Not part of the repo, not committed. Written to capture the full technical rationale that was deliberately kept out of the plugin's user-facing README.md (which was trimmed to stay concise). This is the reference for "why" the code looks the way it does, and a changelog of every review-driven fix applied.

Branch: csafreen/eq5d5l-index-calculation in OHDSI/d2e.

1. Purpose
Computes EQ-5D-5L utility index (health state) scores from questionnaire responses already present in the OMOP observation table, and writes the result into the OMOP measurement table. EQ-5D-5L is a 5-dimension (mobility, self-care, usual activities, pain/discomfort, anxiety/depression), 5-level (1=no problems .. 5=extreme problems) health-related quality-of-life instrument. Each response's 5-digit "health state" code (e.g. 11223) is converted to a single utility index via a country-specific EuroQol value set.

The plugin is a pure consumer in the pipeline: it does not read FHIR data or run any FHIR→OMOP transform. It assumes an upstream EQ5D5L-to-OMOP-Observation FHIR StructureMap has already written 5 observation rows per questionnaire administration.

2. File map
plugins/flows/eq5d5l_index_calculation/
├── pyproject.toml / pixi.lock          # pixi env manifest (no HANA driver - Postgres only)
├── package.json                        # trex flow manifest (parameter_openapi_schema)
└── eq5d5l_index_calculation_plugin/
    ├── flow.py                         # Prefect flow/tasks - the orchestration
    ├── scoring.py                      # pure functions: value-set loading + scoring
    ├── types.py                        # config schema, dimension/concept-id constants
    ├── README.md                       # concise, user-facing docs
    ├── external/value_sets/*.txt       # bundled EuroQol STATA value-set downloads
    └── tests/
        ├── test_flow.py                # 51 tests, mocked DAOs
        └── test_scoring.py             # value-set parsing + scoring unit tests
Shared infrastructure touched by this work (not eq5d5l-specific):

plugins/flows/_shared_flow_utils/dao/sqlalchemydao.py — generic SQLAlchemy-based DAO
plugins/flows/_shared_flow_utils/dao/trexdao.py — TREX/DuckDB pgwire DAO
plugins/flows/_shared_flow_utils/dao/DBDao.py — dialect→DAO-class factory
plugins/flows/data_transformation/dataflow_ui_plugin/nodes.py — FhirMappingNode, DbWriter (used by the drag-and-drop pipeline builder, shared across many pipelines)
3. Functional walkthrough (calculate_eq5d5l_index)
load_value_set(country_code) — resolve the EuroQol value set (see §7).
Dialect guard, then TrexDao-routed dbdao: probe the tenant's underlying dialect via DBDao(database_code=config.database_code).dialect (a credentials lookup only, no connection opened, no cache_id involved) and reject anything but Postgres. Only then build dbdao = DBDao(dialect=TREX, database_code=config.database_code, cache_id=config.omop_dataset_id) — i.e. this dataset's OMOP data is read/written through the Trex cache, not a direct Postgres connection (see §4.1, §6).
read_eq5d5l_observations — select the 5 dimension observation rows via observation_concept_id (fixed mapping in types.py, not configurable per run).
calculate_index_rows — group rows by observation_source_value (the plain QuestionnaireResponse id after the upstream pipeline's stage-2 resolution), score each complete group (5/5 dimensions, agreeing person_id), skip and log incomplete ones. _extract_level() parses value_source_value into an int 1–5, rejecting non-numeric AND out-of-range digits (e.g. "6") — this fix keeps a single bad answer code from aborting the whole flow (it used to reach assemble_health_state() and raise there).
dry_run: return the computed rows without writing anything.
Prerequisite check: _require_fhir_mapping_table against the TREX mapping DAO, checked before the destructive measurement write (see §8) so a missing/outdated prerequisite never leaves measurements replaced with no lineage/metadata to match.
write_measurements — atomic delete-then-insert of this dataset's measurement rows for the EQ-5D-5L measurement_concept_id, via TrexDao.delete_and_insert_rows (see §5.1). Returns the freshly inserted rows and the previous rows it just deleted (both measurement_id and measurement_source_value, not just the id — see §5.2).
write_fhir_key_map — reconcile stale lineage + upsert current lineage (§8).
write_algorithm_metadata — overwrite-on-rerun provenance row in {schema}.metadata (only if inserted_rows is non-empty).
4. Config (Eq5d5lCalculateConfig)
dry_run: bool = False
database_code: str        # tenant connection credentials key
schema_name: str          # OMOP CDM schema of the dataset being scored
omop_dataset_id: str       # the dataset's own id, passed through as cache_id
country_code: str          # ISO 3166-1 alpha-2; selects the EuroQol value set
4.1 database_code vs omop_dataset_id — the saga
These identify genuinely different things:

database_code — the tenant's connection credentials key (e.g. alpdev_pg), looked up in the DATABASE_CREDENTIALS secret by DaoBase.__extract_database_credentials() (daobase.py:519-550), which matches solely on database_code.
omop_dataset_id/cache_id — the dataset's own id. For a real cache/snapshot dataset (built via the portal's createDatasetSnapshot), it's a sanitized UUID (resolveCacheId() / sanitizeIdForCacheId() in plugins/functions/portal/src/dataset/entity/dataset.entity.ts:9-43) — structurally unrelated to database_code. Only "source" (raw tenant connection) dataset rows get cache_id == database_code, by explicit design (confirmed by a portal migration that repaired rows where this had drifted, and by dataset-command.cache-id.test.ts).
Most other flows don't route by cache_id at all. For a direct Postgres/HANA connection, cache_id is not consulted for connection routing — neither SqlAlchemyDao.engine nor IbisDao.ibis_connect() reference it; both resolve via DaoBase's credential lookup, keyed on database_code only. Every flow that reads/writes a dataset's live, current CDM tables directly (phenotype_plugin, cohort_generator_plugin, loyalty_score_plugin, i2b2_plugin) uses DBDao(database_code=..., cache_id=...) with no explicit dialect, and cache_id is inert there.

eq5d5l does not follow that pattern. Per explicit architectural direction, this plugin connects through the Trex cache for both halves of its work, so both the OMOP data and the FHIR lineage are reached the same way — through cache_id-based routing — rather than one going direct-to-Postgres and the other through Trex:

dbdao (this dataset's observation/measurement rows) — DBDao(dialect=TREX, database_code=config.database_code, cache_id=config.omop_dataset_id). cache_id=omop_dataset_id scopes this to this dataset specifically — the same dataset id passed to the flow.
mapping_dao (the tenant-wide FHIR lineage table, fhir_omop_key_map) — DBDao(dialect=TREX, database_code=config.database_code), i.e. no cache_id override, so TrexDao falls back to cache_id = database_code. FHIR lineage is tenant-wide (keyed by database_code in the schema name itself — {database_code}_{schema_name}_fhir_mapping, see §8), not scoped to one dataset, so it is intentionally addressed differently from the per-dataset dbdao.
Both DAOs are TrexDao instances; they differ only in which dbname they connect to (self.cache_id, per TrexDao._get_connection()). This is why TrexDao needed its own select_rows_where_in/delete_and_insert_rows implementations (§5.1) — previously only SqlAlchemyDao had them, which is what limited this plugin to direct-Postgres-only before this change.

Iteration history on this point (for context — this was litigated across several review rounds):

First pass: added a guard rejecting omop_dataset_id != database_code, reasoning that a mismatch meant silently operating on the wrong catalog.
User corrected: database_code and omop_dataset_id are legitimately different concepts and commonly differ in real deployments — the guard would reject the normal case. Reverted.
User further corrected the README's wording and example, which still implied the two values were "usually the same" — fixed the example to show a realistic UUID-shaped omop_dataset_id and explained the real relationship precisely.
A subsequent review round flagged that passing cache_id to a direct-Postgres dbdao was a no-op — the connection never actually used it, so a non-default omop_dataset_id silently read/wrote the wrong (tenant-default) catalog instead of the dataset-scoped one, or was simply misleading dead configuration.
Resolution: rather than rejecting non-default cache_ids (which would break the normal case per point 2) or leaving it a documented no-op, the plugin now routes through TrexDao so cache_id is genuinely load-bearing — this is the dbdao/mapping_dao split described above. This required adding select_rows_where_in/delete_and_insert_rows to TrexDao (§5.1) and replacing the old dialect-allowlist guard with an underlying-dialect probe (§6).
5. Concurrency & transaction safety
5.1 delete_and_insert_rows / select_rows_where_in
sqlalchemydao.py (SqlAlchemyDao, pre-existing, unrelated to this plugin's own calls now — see below). Delete-then-insert in one engine.begin() transaction. When id_column is given (sequential id allocation via MAX(id)+1), the table is locked first (_lock_table_for_id_allocation) via LOCK TABLE ... IN EXCLUSIVE MODE, built from the reflected Table object's own schema-qualified, dialect-quoted name (via connection.dialect.identifier_preparer.format_table(table_obj)) — not a manually concatenated f"{schema}.{table}" string. Supports Postgres and HANA; raises NotImplementedError for any other dialect rather than silently racing. This class and its HANA-specific identifier-casing handling are unchanged and still used by other plugins (see §9) — eq5d5l no longer calls into this class at all now that its dbdao routes through TrexDao instead (§4.1).

trexdao.py (TrexDao, new — added to support the dbdao/mapping_dao TrexDao-routing decision in §4.1). select_rows_where_in() and delete_and_insert_rows() were added to give TrexDao feature parity with SqlAlchemyDao for the two methods this plugin's dbdao needs (read_eq5d5l_observations, write_measurements). Both build their SQL with psycopg2.sql (pg_sql.SQL(...).format(pg_sql.Identifier(...), pg_sql.Literal(...))), matching the parameterized-identifier/literal safety the rest of TrexDao already uses in execute_sql/batch_insert_values, rather than manual string formatting. delete_and_insert_rows(id_column=...) reuses the same LOCK TABLE ... IN EXCLUSIVE MODE + MAX(id)+1 pattern as SqlAlchemyDao, run inside _get_connection(autocommit= False) (§5.3) so the lock, delete, and insert are one transaction. Unlike SqlAlchemyDao, there is no HANA-casing concern here — TrexDao's underlying store (DuckDB via the Trex pgwire cache) is Postgres-identifier-semantics throughout, and eq5d5l is Postgres-only regardless (§6).

5.2 FHIR lineage reconciliation race (write_fhir_key_map)
write_measurements() and write_fhir_key_map() run as separate transactions/tasks. A naive "does this id still exist" check is insufficient: the id allocator can reuse a deleted id for an unrelated row from a concurrent rerun of the same dataset. Two failure modes were identified and fixed:

(a) Writing stale lineage as if current. Fixed by matching on (measurement_id, measurement_source_value) together, not id alone — a same-id row from a concurrent run (different qrId) is correctly recognized as not-this-run's-row and skipped rather than linked.
(b) Deleting a concurrent run's fresh mapping as "stale." write_measurements() now returns not just the previous ids it deleted but the full (measurement_id, measurement_source_value) pairs. _delete_stale_measurement_key_map_rows matches the DELETE on both columns ((omop_id = 'x' AND fhir_id = 'y') OR ...), not omop_id IN (...) alone — so it can never delete a different run's mapping that happens to share a reused id but has a different qrId.
Batching: the OR-predicate is built in _STALE_KEY_MAP_DELETE_BATCH_SIZE (500) chunks — one DELETE statement per batch — so a large rerun's generated SQL stays bounded instead of growing with the whole prior measurement set.
Transaction: the stale-delete batches and the replacement batch_insert_values call now run inside one mapping_dao._get_connection(autocommit=False) block, so a failure between them (e.g. the insert) can't leave the dataset's lineage cleared with nothing written to replace it.
Explicitly reverted per user instruction: an earlier fix had write_fhir_key_map return whether this run's rows were found stale, and gated write_algorithm_metadata on that signal (to stop an older, slower concurrent run from overwriting metadata after a newer run's measurements won). This is a real, still-open race — write_fhir_key_map detects/skips stale rows locally but doesn't report that outcome to the metadata step — but per explicit direction this is not being fixed right now. If revisited, the fix is: have write_fhir_key_map return a boolean/flag, and gate the write_algorithm_metadata call in calculate_eq5d5l_index on it.

5.3 TrexDao._get_connection(autocommit=False)
Added to allow a caller to run several execute_sql()/batch_insert_values() calls (each accepting an optional con=) as one real transaction, instead of each opening its own connection and auto-committing independently. Used by:

write_fhir_key_map (§5.2)
FhirMappingNode._ensure_mapping_schema's index rebuild (§5.4)
FhirMappingNode.task()'s reconcile-then-insert sequence (§8.1)
Startup-statement rollback bug (fixed): when cache_id != database_code, the connection issues a manual USE <cache_id> statement, tolerating failure (e.g. the DuckDB catalog isn't attached yet). In autocommit=False mode, a failed statement leaves the Postgres session's transaction in the "aborted" state — every subsequent statement on that connection then fails with current transaction is aborted, commands ignored until end of transaction block, even though the failure was meant to be tolerated. Fixed by calling con.rollback() immediately after catching the USE failure, but only when not autocommit (under autocommit=True, each statement is already independently committed/failed, so there's no shared transaction to poison).

5.4 FhirMappingNode (nodes.py) — two separate transaction-safety fixes
_ensure_mapping_schema's index rebuild. The old 2-column unique index (fhir_id, fhir_resource_type) is dropped and the new 4-column one (fhir_id, fhir_resource_type, omop_table_name, omop_id) is created as two separate auto-committed statements. If the CREATE fails (e.g. an existing table has duplicates under the wider key), the table is left with no unique index — the old one was already dropped and committed. Fixed by running both statements on one autocommit=False connection, so a failed create rolls back the drop too.
task()'s reconcile-then-insert. Two DELETEs (clearing stale data_source/ fhir_omop_key_map rows for this (omop_table_name, fhir_resource_type)) followed by two batch_insert_values calls were four independently auto-committed operations. A failure between the deletes and the inserts left the dataset's lineage cleared with nothing written back. Fixed the same way: one autocommit=False connection wrapping all four statements, con= threaded through each call.
5.5 DbWriter.task() (nodes.py) — truncate-on-missing-input bug
The upstream-result validation only checked upstream.result is not None and not isinstance(..., pd.DataFrame) — a None result (upstream ran without erroring, but produced no result at all — distinct from a valid, empty full-refresh DataFrame) passed through un-rejected. Since if self.truncate: self._truncate_table(...) runs before the later "empty dataframe, nothing to write" check, a truncate=True writer downstream of a node that returned None would truncate the destination table with nothing to replace it. Fixed by explicitly checking upstream.result is None and rejecting before _truncate_table() is ever reached, while still letting an actual empty DataFrame (a legitimate "full refresh to empty" signal) proceed to truncate.

6. Dialect support
Only Postgres-backed tenants are supported, but the check can't simply read dbdao.dialect — dbdao is now always a TrexDao (§4.1), and TrexDao.dialect is hardcoded to always report "trex" regardless of what's actually behind the cache, so it can't distinguish a Postgres-backed tenant from anything else. Two separate checks run, in order, before any TrexDao-routed work happens:

Underlying-dialect probe. DBDao(database_code=config.database_code).dialect — built with no dialect=/cache_id= kwargs, so DBDao's factory resolves the DAO class from the tenant's own registered dialect (a credentials lookup, per DaoBase.__extract_database_credentials(), §4.1) rather than defaulting to Trex. This answers "what kind of database does this tenant actually have," which dbdao.dialect cannot. Anything other than SupportedDatabaseDialects.POSTGRES raises NotImplementedError naming the actual dialect.
Required-method defense check, only after the dialect probe passes: build the real dbdao = DBDao(dialect=TREX, database_code=..., cache_id=omop_dataset_id) and verify hasattr(dbdao, m) for each of _REQUIRED_DAO_METHODS = ("select_rows_where_in", "delete_and_insert_rows"). Since TrexDao now implements both (§5.1) and dbdao is always a TrexDao, this can't currently fail — it's defense-in-depth against a future change to TrexDao (or to DBDao's registry) silently dropping one of these methods, not a live dialect-routing decision the way it was when dbdao could resolve to different DAO classes.
HANA was explicitly scoped out of this plugin (separate from the shared DAO layer, which does still support HANA — see §5.1). Two real HANA-specific gaps were found along the way and are documented here in case HANA support is revisited:

HANA's OMOP 5.3.1 metadata table (per plugins/flows/data_management/data_management_plugin/db/migrations/hana/changesets/omop/V5.3.1.1.1__apply_v5.3.1.sql:28-37) has no metadata_id or value_as_number column, unlike Postgres's CDM 5.4 shape. write_algorithm_metadata originally unconditionally sent both.
The HANA driver (sqlalchemy-hana + hdbcli) was never declared in this plugin's pyproject.toml, so provision-envs.sh (which only installs it when the manifest contains the sqlalchemy-hana marker) would never provision it for this plugin's env.
Both were fixed at one point and then reverted along with the rest of HANA support once the decision was made to scope this plugin to Postgres only.

7. STATA value-set interpreter (scoring.py)
Rather than pattern-matching a fixed formula shape (an earlier approach that broke the first time a country's real syntax didn't fit the anticipated pattern — Canada has a nonzero level-1 coefficient, per-dimension rather than shared "at level 4/5" indicators, and an additive quadratic correction), scoring.py has a small interpreter (_compile_stata_statements()/_run_stata_statements()) for the actual STATA subset EuroQol's value-set files use: gen/replace <var> = <expr> [if <cond>], arithmetic (+ - * / ^), comparisons (== != >= <= < > <>), booleans (& |), and missing()/round(). It runs the bundled file's real statements once per each of the 3125 possible 5-dimension health states, and records whatever EQ_index ends up with into an index_table.

Fixes applied in this session:

<> (Stata's alternate not-equal spelling) was documented as supported but wasn't actually tokenizable — absent from both the multi-character token alternatives in _STATA_TOKEN_RE and the _STATA_COMPARATORS map. Valid syntax using <> tokenized as two separate </> single-char tokens and failed to parse ("trailing content"). Fixed by adding <> to both.
round() used Python's banker's-rounding instead of Stata's round-half-away-from- zero. round(904.5, 3) in Python rounds to even (904); Stata rounds away from zero (905). The bundled Canada syntax hits this exact tie for health state 11121 (0.9045 before its final round(EQ_index, .001)), so the wrong rounding silently corrupted that one entry. Fixed with a dedicated _stata_round_half_away_from_zero().
health_state_to_index()'s validation only checked "5 digits," not "each digit 1-5." For the stata_simulation (index_table lookup) path this was already safe, since "00000"/"66666" simply aren't keys in the table. For the hand-authored main_effects_interaction path, though, there's no index_table to bound it — levels = [int(c) for c in health_state] would accept 0 or 6 silently (0 has no effect since the code only branches on level >= 2; 6 would look up a coefficient key that doesn't exist). Fixed with an explicit any(c not in "12345" ...) check.
Missing/misspelled main-effect coefficients defaulted to 0.0 silently (coefficients.get(key, 0.0)). A malformed hand-authored value set (typo'd key, or an incomplete transcription) would produce a plausible-looking but clinically incorrect score instead of an error. Fixed to raise ValueError naming the missing key/ dimension/level if the key isn't present — every level 2-5 coefficient must be explicit, including deliberate zeros.
8. FHIR lineage (fhir_omop_key_map)
After writing measurement rows, this plugin upserts one fhir_omop_key_map row per computed index measurement: (fhir_id=qrId, fhir_resource_type="QuestionnaireResponse", omop_table_name="measurement", omop_id=<measurement_id>), into {database_code}_{schema_name}_fhir_mapping.fhir_omop_key_map — mirroring FhirMappingNode. This plugin never creates that schema/table itself; it's a lineage consumer only.

8.1 Prerequisite validation (_require_fhir_mapping_table)
Two checks, both must pass, both checked before the destructive measurement write:

Schema/table existence — if missing, the upstream FHIR→OMOP pipeline for this dataset hasn't run yet.
Index shape, via a new TrexDao.get_indexes_for_table() method (queries DuckDB's duckdb_indexes() table function). A table created by an older FhirMappingNode could still carry the former 2-column unique index rather than the current 4-column one this plugin's ON CONFLICT target requires — existence alone isn't enough, since an outdated index would make the ON CONFLICT (fhir_id, fhir_resource_type, omop_table_name, omop_id) clause fail with "no matching unique constraint" the first time this plugin tried to insert.
Both checks were originally only run inside write_fhir_key_map, after write_measurements()'s destructive delete-and-replace had already committed — meaning a missing/outdated prerequisite left the dataset with new measurement rows and no matching lineage/metadata. Moved to run in calculate_eq5d5l_index, guarded by if rows:, before write_measurements() is ever called.

9. Known limitations / open items
HANA is not supported by this plugin (though the shared DAO layer's HANA handling was preserved/restored, since other plugins depend on it, and is otherwise unchanged by this work — see §5.1).
TrexDao.select_rows_where_in/delete_and_insert_rows are unit-tested only against a stubbed SQL renderer, not a real connection (see §10) — psycopg2.sql.Composable. as_string(context) requires a real psycopg2 connection/cursor (raises TypeError against a MagicMock), so the tests stub .as_string() at the class level rather than exercising psycopg2's actual identifier/literal quoting. Verifying the generated SQL against a real Postgres/Trex/DuckDB connection is an open follow-up.
Metadata-write race (§5.2) is a known, currently-unfixed gap, left as-is per explicit direction: an older, slower concurrent rerun can still overwrite the metadata provenance row after a newer run's measurements have already won the lineage race.
Full per-dataset serialization (locking the whole measurement/lineage/metadata sequence per dataset across concurrent runs) has not been implemented anywhere in this plugin; all concurrency fixes here are "detect and skip/reconcile the specific stale rows," not "prevent the race from happening at all."
10. Testing
test_flow.py (51 tests) mocks all DAOs (MagicMock), exercising calculate_eq5d5l_index end-to-end with a stateful _patch_daos helper that tracks "what was last inserted into measurement" so the lineage re-check logic can be verified against a realistic before/after state. test_scoring.py covers value-set parsing (STATA interpreter, <> operator, rounding tie-breaking, missing-coefficient rejection) and the health_state_to_index/assemble_health_state validation paths directly, without going through the flow layer. No integration tests exist against a real Postgres/HANA/Trex instance — all verification in this session was either unit tests with mocks or small isolated simulations (e.g. SQLite standing in for Postgres/HANA identifier-casing semantics) run manually and discarded, not checked in.

Pre-existing, unrelated bug found and fixed while adding the above: _shared_flow_utils/tests/test_trexdao.py's 5 pre-existing tests all failed with ValueError: Unable to find block document named database-credentials for block type secret. TrexDao.__init__ → DaoBase.__init__ (daobase.py:91) unconditionally calls Secret.load("database-credentials").get() on every DAO construction (even TrexDao, whose actual credentials come from trex-sql-password/Prefect Variables, not that secret) — but the test file's @patch("_shared_flow_utils.dao.trexdao.Secret") only intercepts the Secret name as imported into trexdao.py; it doesn't cover the separate Secret reference daobase.py imports and calls internally. Fixed by adding @patch("_shared_flow_utils.dao.daobase.Secret") to all 9 tests in the file (the 5 pre-existing plus 4 new ones for select_rows_where_in/delete_and_insert_rows).

Fixing this also surfaced the .as_string()/MagicMock issue in the existing test_use_cache_id_issued_when_cache_id_differs_from_database_code test, which called executed_sql.as_string(connect_mock.return_value) directly — once the Secret patch let the test actually reach that line, it failed with the same TypeError described above. Fixed by rendering via Composed.seq/SQL.string/Identifier.string (public, connection-free attributes) instead of .as_string().

All 9 tests in test_trexdao.py pass as of this update.

### Merge Checklist


alicia-koh (approved):

csafreen:
@copilot resolve the merge conflicts in this pull request

suwarnoong (approved):

plugins/flows/eq5d5l_index_calculation/package.json
```diff
@@ -0,0 +1,104 @@
+{
+  "name": "@data2evidence/eq5d5l-index-calculation-flow",
+  "version": "0.0.1",
```
mwaiyee: Should this be `0.0.99` ?

mwaiyee (approved):

PR #3328, opened 2026-09-11
fix(logto): wait for trex provisioning to finish before seeding grants

p-hoffmann:


p-hoffmann:
Withdrawing: this gate closes a dependency cycle rather than ordering the two writers.

logto -> trex health -> server.listen -> atlas-db-init -> logto.users -> logto

In the published d2e-trex:develop image, atlas-db-init still runs before listen, so the cycle is broken only by that step's ~5 minute timeout, at the cost of atlas-db-init being skipped and setupdemo failing at 'Waiting for cache to be ready'.

The signal this needs fires at provisioning-complete (~5s); the health endpoint only answers at boot-complete (~5min), and everything in between includes logto itself. Tuning the timeout would convert the deadlock back into the original race, which the retry in #3327 already handles.

Branch p-hoffmann/gate-logto-on-trex-health retained at 88de14fde for reference.

